PR #2, opened 2021-11-26
add tests - part 1

skshetry:
Migrating https://github.com/iterative/dvc/blob/004e15c2c00a24c7a8fba88611def600a9dd223f/tests/unit/scm/test_git.py.

### DIFF
--- a/scmrepo/git/__init__.py
+++ b/scmrepo/git/__init__.py
@@ -6,7 +6,7 @@ import re
 from collections.abc import Mapping
 from contextlib import contextmanager
 from functools import partialmethod
-from typing import Dict, Iterable, Optional, Tuple, Type
+from typing import Dict, Iterable, Optional, Tuple, Type, Union
 
 from funcy import cached_property, first
 from pathspec.patterns import GitWildMatchPattern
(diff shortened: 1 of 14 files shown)


scmrepo/git/__init__.py
```diff
@@ -260,6 +260,29 @@ def get_fs(self, rev: str):
 
         return GitFileSystem(scm=self, rev=rev)
 
+    @classmethod
+    def init(
```
skshetry, line 264: I implemented `init` and added tests for all backends.

scmrepo/git/__init__.py
```diff
@@ -260,6 +260,29 @@ def get_fs(self, rev: str):
...
+                continue
+            try:
+                backend.init(path, bare=bare)
+                # TODO: reuse created object instead of initializing a new one.
+                return cls(path)
+            except NotImplementedError:
+                pass
+        raise NoGitBackendError("init")
+
+    def add_commit(
```
skshetry, line 278: We don't have `scm_gen`, so I added this `add_commit` to simplify tests. It's questionable, but `scmrepo` should also aim to provide higher-level APIs than underlying backends.

tests/conftest.py
```diff
@@ -0,0 +1,46 @@
...
+    monkeypatch.setenv("GIT_CONFIG_NOSYSTEM", "1")
+    contents = b"""
+[user]
+name=DVC Tester
+email=dvctester@example.com
+[init]
+defaultBranch=master
+"""
+    (home_dir / ".gitconfig").write_bytes(contents)
+    pygit2.settings.search_path[pygit2.GIT_CONFIG_LEVEL_GLOBAL] = str(home_dir)
```
skshetry, line 34: Seems like `pygit2` does not use the monkeypatched environment variable, so we have to set manually.

pmrowla (approved):

Issue #4, opened 2021-12-06
Fix output of `describe` in dulwich backends.


karajan1001:
For issue in dvc: https://github.com/iterative/dvc/issues/6051.

PR #5, opened 2021-12-06
dulwich: fix output of `describe`(#4)

karajan1001:
fix: #4 and https://github.com/iterative/dvc/issues/6051
In dvc we get a random toggling branch name in `exp show`. It is because
when we iter branches the `dulwich.refs.keys` gives an unordered `Set`
of refs. Here we reordered it, and makes current active branch always be
in the first order.

1. Add order to the result of dulwich.refs.keys.
2. Make active branch always in the first place.
3. Add a new test to it.

### DIFF
--- a/scmrepo/git/__init__.py
+++ b/scmrepo/git/__init__.py
@@ -314,7 +314,7 @@ class Git(Base):
     _stash_push = partialmethod(_backend_func, "_stash_push")
     _stash_apply = partialmethod(_backend_func, "_stash_apply")
     _stash_drop = partialmethod(_backend_func, "_stash_drop")
-    describe = partialmethod(_backend_func, "describe")
+    _describe = partialmethod(_backend_func, "_describe")
     diff = partialmethod(_backend_func, "diff")
     reset = partialmethod(_backend_func, "reset")
     checkout_index = partialmethod(_backend_func, "checkout_index")
@@ -397,3 +397,18 @@ class Git(Base):
 
     def _reset(self) -> None:
         self.backends.reset_all()
+
+    def describe(
+        self,
+        rev: str,
+        base: Optional[str] = None,
+        match: Optional[str] = None,
+        exclude: Optional[str] = None,
+    ) -> Optional[str]:
+        if (
+            base == "refs/heads"
+            and self.get_rev() == rev
+            and self.get_ref("HEAD", follow=False).startswith(base)
+        ):
+            return self.get_ref("HEAD", follow=False)
+        return self._describe(rev, base, match, exclude)
--- a/scmrepo/git/backend/dulwich/__init__.py
+++ b/scmrepo/git/backend/dulwich/__init__.py
@@ -590,7 +590,7 @@ class DulwichBackend(BaseGitBackend):  # pylint:disable=abstract-method
         except ValueError as exc:
             raise SCMError("Failed to drop stash entry") from exc
 
-    def describe(
+    def _describe(
         self,
         rev: str,
         base: Optional[str] = None,
--- a/tests/test_git.py
+++ b/tests/test_git.py
@@ -607,3 +607,30 @@ def test_checkout_subdir(tmp_dir: TmpDir, scm: Git, git: Git):
     with (tmp_dir / "dir").chdir():
         git.checkout(rev)
         assert not (tmp_dir / "dir" / "bar").exists()
+
+
+@pytest.mark.skip_git_backend("pygit2", "gitpython")
+def test_describe(tmp_dir: TmpDir, scm: Git, git: Git):
+    tmp_dir.gen({"foo": "foo"})
+    scm.add_commit("foo", message="foo")
+    rev_foo = scm.get_rev()
+
+    tmp_dir.gen({"foo": "bar"})
+    scm.add_commit("foo", message="bar")
+    rev_bar = scm.get_rev()
+
+    assert git.describe(rev_foo, "refs/heads") is None
+
+    scm.checkout("branch", create_new=True)
+    assert git.describe(rev_bar, "refs/heads") == "refs/heads/branch"
+
+    tmp_dir.gen({"foo": "foobar"})
+    scm.add_commit("foo", message="foobar")
+    rev_foobar = scm.get_rev()
+
+    scm.checkout("master")
+    assert git.describe(rev_bar, "refs/heads") == "refs/heads/master"
+    assert git.describe(rev_foobar, "refs/heads") == "refs/heads/branch"
+
+    scm.tag("tag")
+    assert git.describe(rev_bar) == "refs/tags/tag"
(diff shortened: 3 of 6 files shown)


scmrepo/git/backend/dulwich/__init__.py
```diff
@@ -597,16 +597,34 @@ def describe(
...
+            try:
+                current_branch = (
+                    base + "/" + os.fsdecode(active_branch(self.repo))
+                )
+            except (IndexError, ValueError):
+                current_branch = None
+
+        best_ref = None
+
+        for ref in sorted(self.iter_refs(base=base)):
```
pmrowla, line 615: So, one problem with this implementation is that it sorting will be slow in the event that a repo has a lot of refs. `git-describe` doesn't really guarantee any particular ordering - from the documentation it just has to return any ref (with the appropriate base) that matches the requested revision.

> git-describe - Give an object a human readable name based on an available ref

The way that we are setting `best_ref` also requires that we finish the entire iteration before returning a result (so we are even slower in that we sort the refs, and also have to re-iterate over the entire sorted list to find `best_ref`).

Preferring the active branch is fine when `base=refs/heads`, but it would be better for us to just check if the current `HEAD` ref is pointing to a branch, and if so, immediately check if that branch points to `rev` before doing any iteration at all.

Otherwise, we can just fall back to doing the existing unsorted iteration and return the first matching result we find.

pmrowla (changes requested):
Also, this new check for whether or not `HEAD` is a branch and matches `rev` when `base=refs/heads` should maybe be added at the top level `Git.describe()` instead of at the individual backend level (since the order won't be guaranteed by any backend, but we want to at least be consistent for in this particular case regardless of what backend we are using)

So at the top `scmrepo.git.Git.describe()` level we would do this specific check (using the generic `self.get_ref()` calls) and then fall back to returning `self._describe()` (where `_describe = _backend_func("describe")`)

scmrepo/git/backend/dulwich/__init__.py
karajan1001: >  Otherwise, we can just fall back to doing the existing unsorted iteration and return the first matching result we find.

So, OK to give a random branch name if we are using for example `dvc exp show -A` for a current not checkouted one?

pmrowla: In my opinion, yes. It's still up for discussion in the original issue, but in general it's a low priority UI issue.

The active branch/HEAD case is worth addressing specifically because by default we only show HEAD, but in the general case for other revs I would prioritize keeping the current performance over making any further behavior changes to `describe`.

(see also: https://github.com/iterative/dvc/issues/6051#issuecomment-847820247)

scmrepo/git/__init__.py
```diff
@@ -397,3 +397,15 @@ def stash_workspace(self, **kwargs):
...
+    def describe(
+        self,
+        rev: str,
+        base: Optional[str] = None,
+        match: Optional[str] = None,
+        exclude: Optional[str] = None,
+    ) -> Optional[str]:
+        if base == "refs/heads":
+            if self.get_rev() == rev:
+                return base + "/" + self.active_branch()
```
pmrowla, line 410: This won't always work because `HEAD` does always point to a branch (meaning `active_branch()` will raise an exception). (You can test what happens in this case by doing `git checkout` with a tag name or an explicit commit SHA) and then trying to `describe()` that git SHA

karajan1001: Only happened with explicit commit SHA, for tags they have higher priority and not go into `if base == "refs/heads"` branch

scmrepo/git/__init__.py
```diff
@@ -397,3 +397,17 @@ def stash_workspace(self, **kwargs):
...
+    def describe(
+        self,
+        rev: str,
+        base: Optional[str] = None,
+        match: Optional[str] = None,
+        exclude: Optional[str] = None,
+    ) -> Optional[str]:
+        if base == "refs/heads" and self.get_rev() == rev:
+            try:
+                return base + "/" + self.active_branch()
```
pmrowla, line 410: One slightly unrelated issue here is that `active_branch` is currently only implemented in the git-python backend, and ideally we should avoid adding new code in `scmrepo` that depends on git-python.

Really, `active_branch` is also probably a function that shouldn't be implemented via the backends at all (since it can be done at the `Git` level using the generic `get_ref`)

tests/test_dulwich.py
```diff
@@ -26,3 +29,22 @@ def test_dulwich_github_compat(mocker: MockerFixture, algorithm: bytes):
     strings = iter((b"ssh-rsa", key_data))
     packet.get_string = lambda: next(strings)
     _process_public_key_ok_gh(auth, None, None, packet)
+
+
+def test_dulwich_describe(tmp_dir: TmpDir, scm: Git):
```
pmrowla, line 34: I think this test should go in https://github.com/iterative/scmrepo/blob/main/tests/test_git.py (so it can be applied to the other backends when more `_describe` implementations are filled out)

pmrowla: It should probably also test the case where `rev` is a separate branch but not the active one

tests/test_git.py
```diff
@@ -607,3 +607,29 @@ def test_checkout_subdir(tmp_dir: TmpDir, scm: Git, git: Git):
     with (tmp_dir / "dir").chdir():
         git.checkout(rev)
         assert not (tmp_dir / "dir" / "bar").exists()
+
+
+def test_dulwich_describe(tmp_dir: TmpDir, scm: Git):
```
pmrowla, line 612:
```suggestion
@pytest.mark.skip_git_backend("pygit2", "gitpython")
def test_describe(tmp_dir: TmpDir, scm: Git):
```

pmrowla: also, please update the test to use the `git: Git` fixture, and use `git.describe` instead of `scm.describe`

It is safe to use `scm` for calls that set up the git repo and git state in the test function, but the `git` fixture specifically forces using the particular backend we want in the specific call/feature/etc we are testing.

tests/test_git.py
```diff
@@ -607,3 +607,30 @@ def test_checkout_subdir(tmp_dir: TmpDir, scm: Git, git: Git):
...
+def test_describe(tmp_dir: TmpDir, scm: Git, git: Git):
+    tmp_dir.gen({"foo": "foo"})
+    scm.add_commit("foo", message="foo")
+    rev_foo = scm.get_rev()
+
+    tmp_dir.gen({"foo": "bar"})
+    scm.add_commit("foo", message="bar")
+    rev_bar = scm.get_rev()
+
+    assert scm.describe(rev_foo, "refs/heads") is None
```
pmrowla, line 622:
```suggestion
    assert git.describe(rev_foo, "refs/heads") is None
```

All of the `describe` calls in this test (the specific backend behavior we want to test), should use the `git` fixture instead of `scm`.

pmrowla: The other usages of `scm` are fine, but in this case, using `git.<func>` ensures that we are only testing a specific backend's version of `<func>` (and not falling back to an alternate backend implementation)

pmrowla (approved):

PR #10, opened 2021-12-09
`get_ref`: Fix annotated tags

daavoo:
Fixes `get_ref` for annotated tags.

Closes iterative/dvc#6383

From https://github.com/iterative/dvc/pull/7008

### DIFF
--- a/scmrepo/git/backend/pygit2.py
+++ b/scmrepo/git/backend/pygit2.py
@@ -317,13 +317,20 @@ class Pygit2Backend(BaseGitBackend):  # pylint:disable=abstract-method
             )
 
     def get_ref(self, name, follow: bool = True) -> Optional[str]:
-        from pygit2 import GIT_REF_SYMBOLIC
+        from pygit2 import GIT_OBJ_COMMIT, GIT_REF_SYMBOLIC, Tag
 
         ref = self.repo.references.get(name)
         if not ref:
             return None
         if follow and ref.type == GIT_REF_SYMBOLIC:
             ref = ref.resolve()
+        try:
+            obj = self.repo[ref.target]
+            if isinstance(obj, Tag):
+                return str(obj.peel(GIT_OBJ_COMMIT).id)
+        except ValueError:
+            pass
+
         return str(ref.target)
 
     def remove_ref(self, name: str, old_ref: Optional[str] = None):
--- a/tests/test_dulwich.py
+++ b/tests/test_dulwich.py
@@ -1,5 +1,9 @@
 import pytest
 from pytest_mock import MockerFixture
+from pytest_test_utils import TmpDir
+
+from scmrepo.git import Git
+from scmrepo.git.backend.dulwich import DulwichBackend
 
 
 @pytest.mark.parametrize(
@@ -26,3 +30,21 @@ def test_dulwich_github_compat(mocker: MockerFixture, algorithm: bytes):
     strings = iter((b"ssh-rsa", key_data))
     packet.get_string = lambda: next(strings)
     _process_public_key_ok_gh(auth, None, None, packet)
+
+
+def test_dulwich_get_ref(tmp_dir: TmpDir, scm: Git):
+    backend = DulwichBackend(tmp_dir)
+    tmp_dir.gen("foo", "foo")
+    scm.add_commit("foo", message="foo")
+
+    scm.set_ref("refs/exp/base", "refs/heads/master")
+
+    custom_ref = backend.get_ref("refs/exp/base", follow=False)
+    assert str(custom_ref).rstrip("\n") == "refs/heads/master"
+
+    head = scm.get_rev()
+    tag = "my_tag"
+    scm.tag(["-a", tag, "-m", "Annotated Tag"])
+
+    tag_ref = backend.get_ref(f"refs/tags/{tag}", follow=False)
+    assert head == tag_ref
--- a/tests/test_dulwich.py
+++ b/tests/test_dulwich.py
@@ -1,9 +1,5 @@
 import pytest
 from pytest_mock import MockerFixture
-from pytest_test_utils import TmpDir
-
-from scmrepo.git import Git
-from scmrepo.git.backend.dulwich import DulwichBackend
 
 
 @pytest.mark.parametrize(
@@ -30,21 +26,3 @@ def test_dulwich_github_compat(mocker: MockerFixture, algorithm: bytes):
     strings = iter((b"ssh-rsa", key_data))
     packet.get_string = lambda: next(strings)
     _process_public_key_ok_gh(auth, None, None, packet)
-
-
-def test_dulwich_get_ref(tmp_dir: TmpDir, scm: Git):
-    backend = DulwichBackend(tmp_dir)
-    tmp_dir.gen("foo", "foo")
-    scm.add_commit("foo", message="foo")
-
-    scm.set_ref("refs/exp/base", "refs/heads/master")
-
-    custom_ref = backend.get_ref("refs/exp/base", follow=False)
-    assert str(custom_ref).rstrip("\n") == "refs/heads/master"
-
-    head = scm.get_rev()
-    tag = "my_tag"
-    scm.tag(["-a", tag, "-m", "Annotated Tag"])
-
-    tag_ref = backend.get_ref(f"refs/tags/{tag}", follow=False)
-    assert head == tag_ref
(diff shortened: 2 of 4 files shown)


tests/test_dulwich.py
```diff
@@ -26,3 +30,21 @@ def test_dulwich_github_compat(mocker: MockerFixture, algorithm: bytes):
...
+
+def test_dulwich_get_ref(tmp_dir: TmpDir, scm: Git):
+    backend = DulwichBackend(tmp_dir)
+    tmp_dir.gen("foo", "foo")
+    scm.add_commit("foo", message="foo")
+
+    scm.set_ref("refs/exp/base", "refs/heads/master")
+
+    head_ref = backend.get_ref("refs/exp/base", follow=False)
+    assert head_ref.rstrip("\n") == "refs/heads/master"
```
daavoo, line 43: I'm not familiar with dulwich but I think `.rstrip(b"\r\n")` is missing here:

https://github.com/dulwich/dulwich/blob/master/dulwich/refs.py#L825

tests/test_dulwich.py
```diff
@@ -26,3 +30,21 @@ def test_dulwich_github_compat(mocker: MockerFixture, algorithm: bytes):
...
     packet.get_string = lambda: next(strings)
     _process_public_key_ok_gh(auth, None, None, packet)
+
+
+def test_dulwich_get_ref(tmp_dir: TmpDir, scm: Git):
+    backend = DulwichBackend(tmp_dir)
+    tmp_dir.gen("foo", "foo")
+    scm.add_commit("foo", message="foo")
+
+    scm.set_ref("refs/exp/base", "refs/heads/master")
```
pmrowla, line 40:
```suggestion
    scm.set_ref("refs/exp/base", "refs/heads/master", symbolic=True)
```

tests/test_dulwich.py
```diff
@@ -26,3 +30,21 @@ def test_dulwich_github_compat(mocker: MockerFixture, algorithm: bytes):
...
+
+    custom_ref = backend.get_ref("refs/exp/base", follow=False)
+    assert str(custom_ref).rstrip("\n") == "refs/heads/master"
```
pmrowla, line 43: This strip shouldn't be necessary, if `get_ref` is returning a string with whitespace on either end the test should fail.

(I think the newline is probably because of the missing `symbolic` in `set_ref`?)

tests/test_dulwich.py
```diff
@@ -26,3 +30,21 @@ def test_dulwich_github_compat(mocker: MockerFixture, algorithm: bytes):
...
+
+    head = scm.get_rev()
+    tag = "my_tag"
+    scm.tag(["-a", tag, "-m", "Annotated Tag"])
+
+    tag_ref = backend.get_ref(f"refs/tags/{tag}", follow=False)
```
pmrowla, line 49: Is there a reason this is `follow=False`? Tags aren't symbolic refs, and should always be retrieved with `follow=True` (which implies that `get_ref` should return a resolved object SHA and not the raw content of a [symbolic] ref file)

pmrowla (changes requested):

tests/test_dulwich.py
pmrowla: It's returning the newline because the ref created in line 40 is not defined as a symbolic reference.

A symbolic reference file contains a special header (`ref: `) and then the target ref string like:
```
ref: refs/target/ref
```
While a non-symbolic reference file only contains a single 40-byte git SHA like:
```
d670460b4b4aece5915caf5c68d12f560a9fe3e4
```

If you call `set_ref` without the symbolic flag, whatever you pass as the ref target gets written directly to the file (in this case `refs/heads/master` with no symbolic ref header).

When we call `get_ref`, dulwich looks for the symbolic ref header `ref: `, and since that header wasn't created, dulwich treats the it as a regular reference and just returns the first 40 bytes in that file (which would normally be the git SHA and no newline).

tests/test_dulwich.py
```diff
@@ -26,3 +30,21 @@ def test_dulwich_github_compat(mocker: MockerFixture, algorithm: bytes):
     strings = iter((b"ssh-rsa", key_data))
     packet.get_string = lambda: next(strings)
     _process_public_key_ok_gh(auth, None, None, packet)
+
+
+def test_dulwich_get_ref(tmp_dir: TmpDir, scm: Git):
```
pmrowla, line 35: Also, this test really belongs in the main test suite's `test_get_ref`. All 3 of the git backends support `get_ref`, and annotated tag support should not be dulwich specific, so we should test all of them.

Basically, the annotated tag specific part of this test should be added in:
https://github.com/iterative/scmrepo/blob/2561daf0915f5ed616fbaeed6e38c42f111ab354/tests/test_git.py#L205

(the other parts of this test are already implemented in the main `get_ref` test)

tests/test_dulwich.py
daavoo: > Is there a reason this is follow=False?

Just because it is how I reached the bug on DVC side:

- `describe`: https://github.com/iterative/dvc/blob/master/dvc/repo/experiments/show.py#L59
- `get_ref` with `follow=False`: https://github.com/iterative/scmrepo/blob/main/scmrepo/git/backend/dulwich/__init__.py#L607

tests/test_dulwich.py
daavoo: Got it. I just copied from https://github.com/iterative/dvc/blob/master/tests/func/experiments/test_experiments.py#L632 to add a test case replicating the failures on the DVC PR:

https://github.com/iterative/dvc/runs/4302184106?check_suite_focus=true

tests/test_dulwich.py
daavoo: Moved

scmrepo/git/backend/pygit2.py
```diff
@@ -317,13 +317,16 @@ def set_ref(
...
+        from pygit2 import GIT_OBJ_COMMIT, GIT_REF_SYMBOLIC, Tag
 
         ref = self.repo.references.get(name)
         if not ref:
             return None
         if follow and ref.type == GIT_REF_SYMBOLIC:
             ref = ref.resolve()
+        commit, _ = self.repo.resolve_refish(str(ref.target))
+        if isinstance(commit, Tag):
+            return str(commit.peel(GIT_OBJ_COMMIT).id)
```
daavoo, line 329: I added annotated tag to `test_get_ref` and `pygit2` was failing. 
Not sure if this fix makes sene

pmrowla: We shouldn't be using `resolve_refish` in `get_ref`, we've already resolved the ref and have the resolved ID for the object it points to (`ref.target`), so we can get the object directly from `self.repo`. Also, we already have a wrapper for `resolve_refish` that peels tags: https://github.com/iterative/scmrepo/blob/4253fe234d1a26a821356d4121accac11f24a86c/scmrepo/git/backend/pygit2.py#L110

(But we don't want to use `_resolve_refish` as a replacement for `get_ref`, since it will still match when `name` is a branch/tag shortname, or something that looks like a git SHA)

This fix should probably be something like

```suggestion
        obj = self.repo[ref.target]
        if isinstance(obj, Tag):
            return str(obj.peel(GIT_OBJ_COMMIT).id)
```

(this is untested, see the pygit docs for `references` and `objects` https://www.pygit2.org/references.html if it doesn't work)

tests/test_git.py
```diff
@@ -205,7 +205,9 @@ def test_set_ref(tmp_dir: TmpDir, scm: Git, git: Git):
 def test_get_ref(tmp_dir: TmpDir, scm: Git, git: Git):
     tmp_dir.gen({"file": "0"})
     scm.add_commit("file", message="init")
+
     init_rev = scm.get_rev()
+
```
pmrowla, line 210: this isn't a big deal, but try not to add extra whitespace changes for no reason

tests/test_git.py
```diff
@@ -214,9 +216,13 @@ def test_get_ref(tmp_dir: TmpDir, scm: Git, git: Git):
             ): "ref: refs/heads/master",
         }
     )
+    scm.tag(["-a", "annotated", "-m", "Annotated Tag"])
 
     assert init_rev == git.get_ref("refs/foo/bar")
     assert init_rev == git.get_ref("refs/foo/baz")
+    assert init_rev == git.get_ref("refs/tags/annotated")
+    # follow=False to replicate dvc.repo.experiments.show
+    assert scm.get_rev() == git.get_ref("refs/tags/annotated", follow=False)
```
pmrowla, line 225: the comment isn't necessary, we can make it obvious from looking at the test that we are checking the annotated tag behavior for both `possible follow` values

(and we don't need the extra `get_rev()` call)

```suggestion
    assert init_rev == git.get_ref("refs/tags/annotated", follow=True)
    assert init_rev == git.get_ref("refs/tags/annotated", follow=False)
```

pmrowla (changes requested):

pmrowla (approved):

Issue #12, opened 2021-12-13
Add a new function `last_n_commits`


karajan1001:
Add a new function returning a list of commit rev with a depth representing the history of a given ref/commit.

used in https://github.com/iterative/dvc/issues/5489.

PR #13, opened 2021-12-13
Add a new function last_n_commits (#12)

karajan1001:
fix: #12

used in iterative/dvc#5489.

1. Add a new function returning a list of commit rev with a depth
representing the history of a given ref/commit.
2. Implement `gitpython` and `dulwich` backends.
3. Add a new unit test for it.


pmrowla:
I don't think this function is necessary, especially since we already have a similar function in `scmrepo` - `branch_revs`. The last `n` commits from a revision (or ref) can already be retrieved with Git `~` or `^` syntax:

To just get a list of revs which can be passed into DVC functions that take `revs` (and get resolved later):
```py
revs = [f"{rev}~{n}" for n in range(num)]
```

or using `branch_revs` to get the list of git SHAs right away:
```py
revs = list(scm.branch_revs(rev=rev, end_rev=f"{rev}~{num}")
```

Issue #16, opened 2021-12-13
dvc exp pull UnicodeDecodeError when using ssh git backend


kwon-young:
When running `dvc exp pull` with a ssh git remote, I get:

```
☧hnr γmainᵪ dvc exp pull -vvv -j1 -r litis-remote pianosa 'train_gae.num_layers=37'
2021-12-13 13:50:09,697 TRACE: Namespace(cprofile=False, yappi=False, cprofile_dump=None, pdb=False, instrument=False, instrument_open=False, quiet=0, verbose=3, version=None, cd='.', cmd='
pull', force=False, pull_cache=True, dvc_remote='litis-remote', jobs=1, run_cache=False, git_remote='pianosa', experiment='train_gae.num_layers=37', func=<class 'dvc.command.experiments.Cmd
ExperimentsPull'>)
2021-12-13 13:50:09,803 DEBUG: Adding '/home/kchoi/prog/hnr/.dvc/config.local' to gitignore file.
2021-12-13 13:50:09,805 DEBUG: Adding '/home/kchoi/prog/hnr/.dvc/tmp' to gitignore file.
2021-12-13 13:50:10,345 DEBUG: git pull experiment 'pianosa' -> 'refs/exps/80/e4443848e6a61775b67834054b7bbf3f0cb982/train_gae.num_layers=37:refs/exps/80/e4443848e6a61775b67834054b7bbf3f0cb
982/train_gae.num_layers=37'
2021-12-13 13:50:11,016 ERROR: unexpected error - 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)
------------------------------------------------------------
Traceback (most recent call last):
  File "/home/kchoi/mambaforge/envs/hnr/lib/python3.9/site-packages/dvc/main.py", line 55, in main
    ret = cmd.do_run()
  File "/home/kchoi/mambaforge/envs/hnr/lib/python3.9/site-packages/dvc/command/base.py", line 45, in do_run
    return self.run()
  File "/home/kchoi/mambaforge/envs/hnr/lib/python3.9/site-packages/dvc/command/experiments.py", line 758, in run
    self.repo.experiments.pull(
  File "/home/kchoi/mambaforge/envs/hnr/lib/python3.9/site-packages/dvc/repo/experiments/__init__.py", line 1015, in pull
    return pull(self.repo, *args, **kwargs)
[...]

PR #17, opened 2021-12-13
progress: change ascii decoding to utf-8 for localized languages in dulwich backend

kwon-young:
fixes https://github.com/iterative/scmrepo/issues/16

### DIFF
--- a/scmrepo/progress.py
+++ b/scmrepo/progress.py
@@ -1,5 +1,5 @@
-from typing import NamedTuple, Optional, Union
 import locale
+from typing import NamedTuple, Optional, Union
 
 from funcy import compose
 
--- a/scmrepo/progress.py
+++ b/scmrepo/progress.py
@@ -50,7 +50,7 @@ class GitProgressReporter:
 
     def __call__(self, msg: Union[str, bytes]) -> None:
         self._reporter._parse_progress_line(
-            msg.decode("ascii").strip() if isinstance(msg, bytes) else msg
+            msg.decode("utf-8").strip() if isinstance(msg, bytes) else msg
         )
 
     @staticmethod
--- a/scmrepo/progress.py
+++ b/scmrepo/progress.py
@@ -1,4 +1,5 @@
 from typing import NamedTuple, Optional, Union
+import locale
 
 from funcy import compose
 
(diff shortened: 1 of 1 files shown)


scmrepo/progress.py
```diff
@@ -50,7 +50,7 @@ def __init__(self, fn) -> None:
 
     def __call__(self, msg: Union[str, bytes]) -> None:
         self._reporter._parse_progress_line(
-            msg.decode("ascii").strip() if isinstance(msg, bytes) else msg
+            msg.decode("utf-8").strip() if isinstance(msg, bytes) else msg
```
skshetry, line 53: hmm, probably we should use [`locale.getpreferredencoding()`](https://docs.python.org/3/library/locale.html#locale.getpreferredencoding) here? Let me look at dulwich.

kwon-young: On my system, `locale.getpreferredencoding()` returns 'UTF-8`.
I'll change the code to use that.

skshetry:
@kwon-young, could you please do the following:
```python
pip install nox

nox -s lint
```

And, add/commit the changes, please. This project is very young, so we don't have contributing guidelines, really sorry for the inconvenience.

kwon-young:
done

scmrepo/progress.py
pmrowla: The message comes from the server, and depends on the server encoding: https://github.com/dulwich/dulwich/issues/687#issuecomment-462747144

For the purposes of this PR I think we should just assume that any typical server implementation is going to be returning `utf-8` (the local environment's `locale.getpreferredencoding()` is not what we want here) @kwon-young 

In the future we should consider how to handle not being able to decode the progress messages (and maybe just display some generic message) @skshetry

skshetry: @pmrowla, we could use `.decode(errors=error)` with `replace` or `surrogateescape`.

kwon-young: > The message comes from the server, and depends on the server encoding: dulwich/dulwich#687 (comment)

Ho I did not know that.

I'll change back to utf-8 with `errors='replace'` (I suppose `surrogateescape` does not really matter since it is only a progress message)

skshetry (approved):
Thanks, @kwon-young for the contributions. We have tests missing for the progress which I'll handle later.

PR #24, opened 2021-12-28
fs: detach from local filesystem

efiop:
This PR makes our fs accept paths relative to the root of the repo and using / as a separator.
Previous behaviour was a legacy from old dvc.

Pre-requisite for making dvcfs/repofs/etc switch to the same relative repo paths and / separators.

### DIFF
--- a/scmrepo/fs.py
+++ b/scmrepo/fs.py
@@ -1,5 +1,6 @@
 import errno
 import os
+import posixpath
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -31,7 +32,6 @@ def bytesio_len(obj: "BytesIO") -> Optional[int]:
 
 class GitFileSystem(AbstractFileSystem):
     # pylint: disable=abstract-method
-    sep = os.sep
     cachable = False
 
     def __init__(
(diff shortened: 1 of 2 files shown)


scmrepo/fs.py
```diff
@@ -145,15 +137,13 @@ def walk(  # pylint: disable=arguments-differ
...
+            root = self.sep.join(prefix) if prefix else ""
 
-            if prefix:
-                root = os.path.join(root, os.sep.join(prefix))
             if detail:
                 yield (
                     root,
-                    {d: self.info(os.path.join(root, d)) for d in dirs},
-                    {f: self.info(os.path.join(root, f)) for f in files},
+                    {d: self.info(posixpath.join(root, d)) for d in dirs},
```
efiop, line 145: For the record: Will switch to `fs.path` once it is introduced to fsspec.

scmrepo/fs.py
```diff
@@ -53,20 +53,12 @@ def __init__(
...
     def _get_key(self, path: str) -> Tuple[str, ...]:
-        from scmrepo.utils import relpath
-
-        if os.path.isabs(path):
-            path = relpath(path, self.root_dir)
-        relparts = path.split(os.sep)
+        relparts = path.split(self.sep)
         if relparts == ["."]:
             return ()
         return tuple(relparts)
```
efiop, line 64: Will introduce `_strip_protocol` later.

PR #26, opened 2022-01-10
pygit2: fix posixpath conversion from windows on is_ignored

### LINKED ISSUE #11: `test_ignored` is failing on `pygit2` backend on Windows.
Need to investigate why this is failing on `pygit2` backend on Windows.

https://github.com/iterative/scmrepo/blob/4253fe234d1a26a821356d4121accac11f24a86c/tests/test_git.py#L716-L718

_Originally posted by @skshetry in https://github.com/iterative/scmrepo/pull/8#r766488463_

skshetry:
When we were converting windows path to posixpath, the values
were not in effect because we were not setting it to the
variable.

Also fixes #11, which unskips  on Windows + pygit2 backend, which
was due to the above bug.


skshetry:
Test failures are unrelated, see #27.

efiop (approved):
Great catch! 🔥

PR #29, opened 2022-01-12
dulwich/pygit: implement backend clone

pmrowla:
- Moves `Git.clone(rev=...)` checkout handling out of backend `clone()` and into `Git.clone()`. Separating this behavior allows us to use `clone()` from one backend and `checkout()` from another, since some backends may implement one but not the other.
- Implements clone in dulwich backend ~~(pending fix for https://github.com/dulwich/dulwich/issues/920)~~
- Implements clone in pygit backend (shallow clone not supported)

Will close https://github.com/iterative/dvc/issues/2215

### DIFF
--- a/scmrepo/git/__init__.py
+++ b/scmrepo/git/__init__.py
@@ -115,11 +115,20 @@ class Git(Base):
         return Stash(self)
 
     @classmethod
-    def clone(cls, url, to_path, **kwargs):
+    def clone(
+        cls,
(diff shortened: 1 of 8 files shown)


pmrowla:
Dulwich `clone()` does not currently have progress support (see https://github.com/dulwich/dulwich/pull/929) but everything else works in the latest release (`0.20.32`)

Issue #30, opened 2022-01-28
`pygit2` backend didn't raise a proper exception in `resolve_rev`.


karajan1001:
How to reproduce it:
```bash
$ mkdir scmtest
$ cd scmtest
# this is correct
$ python3 -c "from scmrepo.git import Git; Git('.').pygit2.resolve_rev('HEAD~')"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/gao/anaconda3/envs/dvc/lib/python3.8/site-packages/scmrepo/git/backend/pygit2.py", line 271, in resolve_rev
    raise RevError(f"unknown Git revision '{rev}'")
scmrepo.exceptions.RevError: unknown Git revision 'HEAD~'

# add some remote
$ git remote add origin someremote
# do not wrap the exception properly from `pygit2`
$ python3 -c "from scmrepo.git import Git; Git('.').pygit2.resolve_rev('HEAD~')"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/gao/anaconda3/envs/dvc/lib/python3.8/site-packages/scmrepo/git/backend/pygit2.py", line 263, in resolve_rev
    shas = {
  File "/Users/gao/anaconda3/envs/dvc/lib/python3.8/site-packages/scmrepo/git/backend/pygit2.py", line 264, in <setcomp>
    self.get_ref(f"refs/remotes/{remote.name}/{rev}")
  File "/Users/gao/anaconda3/envs/dvc/lib/python3.8/site-packages/scmrepo/git/backend/pygit2.py", line 322, in get_ref
    ref = self.repo.references.get(name)
  File "/Users/gao/anaconda3/envs/dvc/lib/python3.8/site-packages/pygit2/repository.py", line 1440, in get
    return self[key]
  File "/Users/gao/anaconda3/envs/dvc/lib/python3.8/site-packages/pygit2/repository.py", line 1436, in __getitem__
    return self._repository.lookup_reference(name)
_pygit2.InvalidSpecError: refs/remotes/origin/HEAD~: the given reference name 'refs/remotes/origin/HEAD~' is not valid

# gitpython works correctly
python3 -c "from scmrepo.git import Git; Git('.').gitpython.resolve_rev('HEAD~')"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
[...]

PR #32, opened 2022-02-07
pygit2: resolve_rev raise a proper exception with `~<n>` like referen…

karajan1001:
fix: #30
Current pygit2 backend's resolve_rev didn't raise a proper exception when dealing with
`HEAD~n` like references.

1. Catch the exception in resolve_rev and reraise it into a proper format.
2. Add a test for it.

In `resolve_refish` both `HEAD~n` like references and `non-exist` references raise a ValueError, can't distinguish them. And I don't think we should let Inner exception `InvalidSpecError` be thrown outside of `scmrepo`. But this method is a relatively slow one, because we need to iter over all of the remotes.

### DIFF
--- a/scmrepo/git/backend/pygit2.py
+++ b/scmrepo/git/backend/pygit2.py
@@ -317,9 +317,17 @@ class Pygit2Backend(BaseGitBackend):  # pylint:disable=abstract-method
             )
 
     def get_ref(self, name, follow: bool = True) -> Optional[str]:
-        from pygit2 import GIT_OBJ_COMMIT, GIT_REF_SYMBOLIC, Tag
+        from pygit2 import (
+            GIT_OBJ_COMMIT,
+            GIT_REF_SYMBOLIC,
+            InvalidSpecError,
+            Tag,
+        )
 
-        ref = self.repo.references.get(name)
+        try:
+            ref = self.repo.references.get(name)
+        except InvalidSpecError:
+            return None
         if not ref:
             return None
         if follow and ref.type == GIT_REF_SYMBOLIC:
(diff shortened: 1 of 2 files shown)


pmrowla:
I'm still confused as to why this PR is needed, since `exp show -n ...` and pygit `resolve_rev(f"HEAD~{n}")` worked prior to the recent refactors. It seems like this might need some more investigation into why `resolve_rev` started failing here (since the exception was not raised at all before)

karajan1001:
> I'm still confused as to why this PR is needed, since `exp show -n ...` and pygit `resolve_rev(f"HEAD~{n}")` worked prior to the recent refactors. It seems like this might need some more investigation into why `resolve_rev` started failing here (since the exception was not raised at all before)
> 
> Is the issue that the exception is raised when `n` is invalid (and that relative commit does not actually exist)? Or that it is still raised when `HEAD~n` points to a valid commit (and that the problem is with using `~` at all)?

It raises when `n` is invalid, I add two test cases here.

```
    assert git.resolve_rev("HEAD~1") == init_rev # this is valid , and works pretty fine in previous.

    with pytest.raises(RevError):
        git.resolve_rev("HEAD~3") # in this case, the previous version will raise `InvalidSpecError` instead of `SCMError`
````

In the previous version of DVC, if we input an invalid number of n, it will raise an unexcepted error. 

```
$ dvc exp show -n 100 --no-pager
ERROR: unexpected error - refs/remotes/origin/HEAD~11: the given reference name 'refs/remotes/origin/HEAD~11' is not valid

Having any troubles? Hit us up at https://dvc.org/support, we are always happy to help!
```

pmrowla:
@karajan1001 this needs a rebase for the CI fixes

karajan1001:
@mergifyio rebase

pmrowla (approved):

PR #45, opened 2022-03-16
status: add warning on slow status call

dtrifiro:
While playing with `dvc exp`, I noticed that it would take a huge amount of time before getting any output.

I tracked this down to [`dvc/repo/experiments/__init__.py#L401`](https://github.com/iterative/dvc/blob/cd65400d83df9ede8f25317a413fe2da5ba05ff3/dvc/repo/experiments/__init__.py#L401): a `self.scm.status()` was taking a long time because of a large untracked directory (`.venv`  virtualenv folder).

This is a kinda naive patch that prints a warning whenever it takes more than 10s for status to return.  I don't intend for this to be merged, rather to be a starting point for a discussion.

### DIFF
--- a/scmrepo/git/__init__.py
+++ b/scmrepo/git/__init__.py
@@ -3,10 +3,11 @@
 import logging
 import os
 import re
+import threading
 from collections.abc import Mapping
 from contextlib import contextmanager
 from functools import partialmethod
-from typing import Dict, Iterable, Optional, Tuple, Type, Union
+from typing import Any, Dict, Iterable, Optional, Tuple, Type, Union
 
 from funcy import cached_property, first
 from pathspec.patterns import GitWildMatchPattern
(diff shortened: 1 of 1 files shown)


dtrifiro:
See https://github.com/iterative/dvc/issues/7471

pmrowla:
We probably should just submit an upstream patch to optimize dulwich `porcelain.status` with an option to skip walking untracked directories on status (similar to the existing option to skip walking ignored directories)

https://github.com/jelmer/dulwich/blob/0aa42073297748b07f5d176557b2357e87ddffbe/dulwich/porcelain.py#L1218

dtrifiro:
Good point, I'll have a look

dtrifiro:
Opened https://github.com/jelmer/dulwich/pull/956, will open a new one here when/if that is merged

dtrifiro:
Getting rid of this, ended up implenting Peter's suggestion in #69

PR #49, opened 2022-03-22
Solve `pygit2` version conflict

karajan1001:
In version 1.5.0, I got a 
> ImportError: cannot import name 'GIT_MERGE_PREFERENCE_FASTFORWARD_ONLY' from 'pygit2'

`GIT_MERGE_PREFERENCE_FASTFORWARD_ONLY` was introduced in 1.6.0. Upgrade the min version of `pygit2` to fix it.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ packages = find:
 install_requires=
     gitpython>3
     dulwich>=0.20.34
-    pygit2>=1.5.0
+    pygit2>=1.6.0
     pygtrie>=2.3.2
     fsspec>=2021.7.0
     pathspec>=0.9.0,<0.10.0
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -26,7 +26,7 @@ packages = find:
 install_requires=
     gitpython>3
     dulwich>=0.20.34
-    pygit2>=1.5.0
+    pygit2>=1.6.0
```
pmrowla, line 29: I think we should make this 1.7.2 (see https://github.com/iterative/scmrepo/pull/50)

pmrowla (approved):

skshetry:
@karajan1001, what do you think of releasing a new version? I suspect https://github.com/iterative/dvc/issues/7498 is similar to this and may be fixed by it.

karajan1001:
@skshetry 
You can release it for now, although I have something else to update, but haven't finished for now.

PR #51, opened 2022-03-28
git: reuse most recent backend when possible

pmrowla:
- When selecting a backend to use for a given git call, we will now try the most recently used backend first, before trying the remaining backends
- If we are switching backends, close the previously used backend before making the git call w/new backend

Should fix https://github.com/iterative/dvc/issues/7458

### DIFF
--- a/scmrepo/git/__init__.py
+++ b/scmrepo/git/__init__.py
@@ -3,6 +3,7 @@
 import logging
 import os
 import re
+from collections import OrderedDict
 from collections.abc import Mapping
 from contextlib import contextmanager
 from functools import partialmethod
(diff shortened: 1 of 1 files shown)


scmrepo/git/__init__.py
```diff
@@ -71,6 +77,10 @@ def reset_all(self) -> None:
         for backend in self.initialized.values():
             backend._reset()  # pylint: disable=protected-access
 
+    def move_to_end(self, key: str, last: bool = True):
```
pmrowla, line 80: We could just make this `move_to_start()` and get rid of the `last` option (since that's all we will ever use), but I went with this for now to stay consistent with the `OrderedDict` call

PR #54, opened 2022-04-07
gitfs: ls: fix return value if detail=True

efiop:
fsspec requires it to return a list of infos, not a dict.

### DIFF
--- a/scmrepo/fs.py
+++ b/scmrepo/fs.py
@@ -1,5 +1,6 @@
 import errno
 import os
+from itertools import chain
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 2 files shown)


efiop:
For the record: tested with dvc, so this is safe to release 🙂

scmrepo/fs.py
```diff
@@ -160,6 +161,6 @@ def walk(  # pylint: disable=arguments-differ
 
     def ls(self, path, detail=True, **kwargs):
         for _, dirs, files in self.walk(path, detail=detail, **kwargs):
-            merge = files.update if detail else files.extend
-            merge(dirs)
-            return files
```
efiop, line 165: Updating files return from walk in place is not as dangerous as as updating `dirs`(because it will change where walk proceeds walking into), but it is still not nice 🙂

PR #55, opened 2022-04-08
dulwich: account for partial pkt reads in asyncssh vendor

pmrowla:
Dulwich expects the underlying SSH vendor to handle partial reads, when dulwich calls `read` with a specific length `n`, it expects the git server to return exactly that many bytes (it does not use `n` as a generic max read buffer size). In our asyncssh vendor if dulwich requests `n` bytes, we need to continue reading until the server has sent all the expected data.

Will fix https://github.com/iterative/dvc/issues/7361
Will fix https://github.com/iterative/dvc/issues/7537

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,9 +44,11 @@ dev =
     # https://github.com/docker/docker-py/issues/2902
     pytest-docker==0.10.3; python_version < '3.10' or sys_platform != 'win32'
     pylint==2.11.1
+    mock==4.0.3
     mypy==0.910
     paramiko==2.8.1
     types-certifi==2021.10.8.0
+    types-mock==4.0.13
     types-paramiko==2.8.1
 
 [options.packages.find]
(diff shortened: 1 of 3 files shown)


tests/test_dulwich.py
```diff
@@ -107,8 +110,37 @@ def test_run_command_data_transfer(server: Server, ssh_port: int):
...
+        username=USER,
+        port=ssh_port,
+        password=PASSWORD,
+    )
+
+    mock_stdout = mocker.patch.object(
+        con.proc.stdout,
+        "read",
+        side_effect=[b"s", b"tdout", b"\n"],
+        new_callable=mocker.AsyncMock,
```
pmrowla, line 131: `patch()` automatically uses `AsyncMock` in Python 3.8+, but not 3.7

setup.cfg
```diff
@@ -44,6 +44,7 @@ dev =
     # https://github.com/docker/docker-py/issues/2902
     pytest-docker==0.10.3; python_version < '3.10' or sys_platform != 'win32'
     pylint==2.11.1
+    mock>=4.0.0
```
pmrowla, line 47: `mock` is needed for backport of `AsyncMock` in py3.7 (only available in the stdlib `unittest.mock` in 3.8+)

PR #40, opened 2022-03-07
path changes from #24 and #25

skshetry:
Reverts iterative/scmrepo#39

PR #56, opened 2022-04-11
gitfs: better comply with fsspec

efiop:
Closes #40

### DIFF
--- a/scmrepo/fs.py
+++ b/scmrepo/fs.py
@@ -20,6 +20,124 @@ if TYPE_CHECKING:
     from scmrepo.git.objects import GitTrie
 
 
+class Path:
+    def __init__(self, sep, getcwd=None, realpath=None):
+        def _getcwd():
+            return ""
+
+        self.getcwd = getcwd or _getcwd
+        self.realpath = realpath or self.abspath
+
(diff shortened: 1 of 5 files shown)


scmrepo/fs.py
```diff
@@ -54,22 +52,14 @@ def __init__(
...
     def _get_key(self, path: str) -> Tuple[str, ...]:
-        from scmrepo.utils import relpath
-
-        if os.path.isabs(path):
-            path = relpath(path, self.root_dir)
-        relparts = path.split(os.sep)
-        if relparts == ["."]:
-            return ()
+        relparts = path.split(self.sep)
+        if relparts and relparts[0] in (".", ""):
```
efiop, line 61: Much better to normalize the path beforehand, but our `fs.path` is not yet available in fsspec, so we'll get to it later.

scmrepo/fs.py
```diff
@@ -20,6 +20,124 @@
     from scmrepo.git.objects import GitTrie
 
 
+class Path:
```
efiop, line 23: Temporary copy from dvc/fs/path.py until we get it into fsspec. Useful for handling chdirs.

Issue #58, opened 2022-04-15
Update on `push_refspec` and `fetch_refspec`


karajan1001:
1. Unify the API of `push_refspec` and  `fetch_refspec`.
2. Can push multi `refspec` for one time.
3. No repository Error handle for `fetch_refspec`
4. Now the `push_refspec` and  `fetch_refspec` will return the status for each refspec.

PR #59, opened 2022-04-15
Update on push_refspec and fetch_refspec (#58)

karajan1001:
fix: #58

1. Unify the API of `push_refspec` and  `fetch_refspec`.
2. Can push multi `refspec` for one time.
3. No repository Error handle for `fetch_refspec`
4. Now the `push_refspec` and  `fetch_refspec` will return the status for each refspec.

### DIFF
--- a/scmrepo/git/__init__.py
+++ b/scmrepo/git/__init__.py
@@ -348,7 +348,7 @@ class Git(Base):
     iter_refs = partialmethod(_backend_func, "iter_refs")
     iter_remote_refs = partialmethod(_backend_func, "iter_remote_refs")
     get_refs_containing = partialmethod(_backend_func, "get_refs_containing")
-    push_refspec = partialmethod(_backend_func, "push_refspec")
+    push_refspecs = partialmethod(_backend_func, "push_refspecs")
     fetch_refspecs = partialmethod(_backend_func, "fetch_refspecs")
     _stash_iter = partialmethod(_backend_func, "_stash_iter")
     _stash_push = partialmethod(_backend_func, "_stash_push")
--- a/scmrepo/git/backend/base.py
+++ b/scmrepo/git/backend/base.py
@@ -28,7 +28,7 @@ class NoGitBackendError(SCMError):
 
 class SyncStatus(Enum):
     SUCCESS = 0
-    DUPLICATED = 1
+    UP_TO_DATE = 1
     DIVERGED = 2
 
 
@@ -226,7 +226,7 @@ class BaseGitBackend(ABC):
 
         Args:
             url: Git remote name or absolute Git URL.
-            refspecs: Iterable containing refspecs to fetch.
+            refspecs: Iterable containing refspecs to push.
                 Note that this will not match subkeys.
             force: If True, remote refs will be overwritten.
             on_diverged: Callback function which will be called if local ref
--- a/scmrepo/git/backend/base.py
+++ b/scmrepo/git/backend/base.py
@@ -1,5 +1,6 @@
 import os
 from abc import ABC, abstractmethod
+from enum import Enum
 from typing import (
     TYPE_CHECKING,
     Callable,
@@ -23,6 +24,12 @@ if TYPE_CHECKING:
 class NoGitBackendError(SCMError):
     def __init__(self, func):
         super().__init__(f"No valid Git backend for '{func}'")
+
+
+class SyncStatus(Enum):
+    SUCCESS = 0
+    DUPLICATED = 1
+    DIVERGED = 2
 
 
 class BaseGitBackend(ABC):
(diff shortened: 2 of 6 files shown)


scmrepo/git/backend/dulwich/__init__.py
```diff
@@ -37,6 +38,13 @@
 logger = logging.getLogger(__name__)
 
 
+class SyncStatus(Enum):
```
karajan1001, line 41: Not sure if there are better solutions for returning the status of push/pull

scmrepo/git/backend/dulwich/__init__.py
```diff
@@ -491,23 +499,20 @@ def get_refs_containing(self, rev: str, pattern: Optional[str] = None):
     def push_refspec(
         self,
         url: str,
-        src: Optional[str],
-        dest: str,
+        refspecs: Union[str, Iterable[str]],
```
pmrowla, line 502: the function signature needs to be updated in `backend/base.py` and the other backends

scmrepo/git/backend/dulwich/__init__.py
```diff
@@ -491,23 +499,20 @@ def get_refs_containing(self, rev: str, pattern: Optional[str] = None):
...
         force: bool = False,
-        on_diverged: Optional[Callable[[str, str], bool]] = None,
         progress: Callable[["GitProgressEvent"], None] = None,
         **kwargs,
-    ):
+    ) -> Mapping[str, int]:
```
pmrowla, line 506: `SyncStatus` should also be defined in `base.py`, and this should be `Mapping[str, SyncStatus]`

scmrepo/git/backend/base.py
```diff
@@ -251,11 +247,6 @@ def fetch_refspecs(
             refspecs: Iterable containing refspecs to fetch.
                 Note that this will not match subkeys.
             force: If True, local refs will be overwritten.
-            on_diverged: Callback function which will be called if local ref
-                and remote have diverged and force is False. If the callback
-                returns True the local ref will be overwritten.
-                Callback will be of the form:
-                    on_diverged(local_refname, remote_sha)
```
pmrowla, line 258: What's the reasoning for removing `on_diverged` support?

karajan1001: In previous, `on_diverged` was only used for raising Exceptions when diverged. Because for now, we will not stop and raise this exception even if they diverged from remote. And for the force push, we already have a `force` flag for this.

karajan1001: Do you have any further usage for this?

karajan1001: Noticed that there was some related code in `executor` do you have any thoughts on this?

pmrowla: It's a useful callback to support because the desired behavior may be to keep the existing one, without considering it a failing state. Passing the refs through the callback allows the caller to do an actual diff on the diverged commits, and decide whether or not they should force push/pull, do nothing, or do a merge/rebase as needed.

We don't use it right now in DVC, but I think we should keep support for the callback in scmrepo

pmrowla: Basically with scmrepo, we should provide useful general API wrappers around the git backends, even if we don't use them in DVC

scmrepo/git/backend/base.py
```diff
@@ -206,25 +213,21 @@ def get_refs_containing(self, rev: str, pattern: Optional[str] = None):
...
         """Push refspec to a remote Git repo.
 
         Args:
             url: Git remote name or absolute Git URL.
-            src: Local refspec. If src ends with "/" it will be treated as a
-                prefix, and all refs inside src will be pushed using dest
-                as destination refspec prefix. If src is None, dest will be
-                deleted from the remote.
-            dest: Remote refspec.
+            refspecs: Iterable containing refspecs to fetch.
```
pmrowla, line 229:
```suggestion
            refspecs: Iterable containing refspecs to push.
```

scmrepo/git/backend/base.py
```diff
@@ -25,6 +26,12 @@ def __init__(self, func):
         super().__init__(f"No valid Git backend for '{func}'")
 
 
+class SyncStatus(Enum):
+    SUCCESS = 0
+    DUPLICATED = 1
```
pmrowla, line 31:
```suggestion
    UP_TO_DATE = 1
```

I think it's better to use the same terms as git's UI here

pmrowla (changes requested):
A couple minor fixes but otherwise LGTM

pmrowla (approved):

efiop:
There seem to be different names in scmrepo and in dvc. We have push_refspecs here, while we use push_refspec (without s) at the end in dvc.

scmrepo/git/__init__.py
```diff
@@ -348,7 +348,7 @@ def add_commit(
     iter_refs = partialmethod(_backend_func, "iter_refs")
     iter_remote_refs = partialmethod(_backend_func, "iter_remote_refs")
     get_refs_containing = partialmethod(_backend_func, "get_refs_containing")
-    push_refspec = partialmethod(_backend_func, "push_refspec")
+    push_refspecs = partialmethod(_backend_func, "push_refspecs")
```
efiop, line 351: @karajan1001 @pmrowla Is renaming intentional? I see that we use `push_refspec` in dvc, so this release seems to break experiments in dvc

efiop: Ah, ok, just need to update dvc to work with 0.0.19 (we have it pinned in dvc to 0.0.18), no biggie. Sorry for the noise.

efiop: Correction: I just didn't rebase in awhile and didn't notice that it was fixed already quite some time ago. Thank you :)

PR #61, opened 2022-05-02
fs, pygit2: add missing f-string formatter for error messages

dtrifiro:


dtrifiro:
CI failures are unrelated, it seems the `six` dependency is not properly resolved on windows on 3.10

efiop (approved):

Issue #65, opened 2022-05-16
linting: type checking failures with dulwich 0.20.36


dtrifiro:
There have been some [type checking related changes](https://github.com/jelmer/dulwich/compare/upstream/0.20.35...dulwich-0.20.36) in the latest dulwich release ([0.20.36](https://github.com/jelmer/dulwich/releases/tag/dulwich-0.20.36)) which break mypy type checking:

```console
nox > python -m mypy
scmrepo/git/backend/dulwich/__init__.py: note: In member "fetch_refspecs" of class "DulwichBackend":
scmrepo/git/backend/dulwich/__init__.py:630:33: error: Argument
"determine_wants" to "fetch" of "GitClient" has incompatible type
"Callable[[Any], Any]"; expected
"Optional[Callable[[Dict[bytes, bytes], Optional[int]], List[bytes]]]" 
[arg-type]
                    determine_wants=determine_wants,
                                    ^
Found 1 error in 1 file (checked 26 source files)
nox > Command python -m mypy failed with exit code 1
nox > Session lint failed.
```

PR #66, opened 2022-05-17
tests: add iter-remote-refs test

dtrifiro:
- add test for `iter-remote-refs` refs (broken for `dulwich<0.20.38`)
- bump dulwich version to `>=0.20.38`

Related https://github.com/iterative/dvc/issues/7670

### DIFF
--- a/tests/test_git.py
+++ b/tests/test_git.py
@@ -9,7 +9,12 @@ from git import Repo as GitPythonRepo
 from pytest_test_utils import TempDirFactory, TmpDir
 from pytest_test_utils.matchers import Matcher
 
-from scmrepo.exceptions import MergeConflictError, RevError, SCMError
+from scmrepo.exceptions import (
+    InvalidRemote,
+    MergeConflictError,
+    RevError,
+    SCMError,
+)
 from scmrepo.git import Git
 
 # pylint: disable=redefined-outer-name,unused-argument,protected-access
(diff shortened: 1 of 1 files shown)


tests/test_git.py
```diff
@@ -417,6 +422,49 @@ def test_fetch_refspecs(
     assert baz_rev == scm.get_ref("refs/foo/baz")
 
 
+@pytest.mark.skip_git_backend("pygit2", "gitpython")
+@pytest.mark.parametrize("use_url", [True, False])
+@pytest.mark.xfail(
+    os.name == "nt",
+    reason="file:// urls are not supported by dulwich on windows",
```
dtrifiro, line 429: This xfail can be removed when https://github.com/jelmer/dulwich/pull/965 is merged and dulwich is upgraded

dtrifiro:
CI failures are related to linting, see #65

tests/test_git.py
pmrowla: looks like this should be fixed in the 0.20.38 which was just released?

dtrifiro: Indeed, let me get rid of this, but before that, we need to get rid of the `file://` urls workaround for windows

dtrifiro: https://github.com/iterative/scmrepo/pull/68

pmrowla (approved):

PR #67, opened 2022-05-17
dulwich: update fetch_refspecs for dulwich 0.20.37

### LINKED ISSUE #65: linting: type checking failures with dulwich 0.20.36

dtrifiro:
dulwich 0.20.37 updated type hints for `fetch` (see https://github.com/jelmer/dulwich/pull/957)

fixes #65

### DIFF
--- a/scmrepo/git/backend/dulwich/__init__.py
+++ b/scmrepo/git/backend/dulwich/__init__.py
@@ -581,7 +581,7 @@ class DulwichBackend(BaseGitBackend):  # pylint:disable=abstract-method
         self,
         url: str,
         refspecs: Union[str, Iterable[str]],
-        force: Optional[bool] = False,
+        force: bool = False,
         on_diverged: Optional[Callable[[str, str], bool]] = None,
         progress: Callable[["GitProgressEvent"], None] = None,
         **kwargs,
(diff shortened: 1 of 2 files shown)


scmrepo/git/backend/base.py
```diff
@@ -240,7 +241,7 @@ def push_refspecs(
     def fetch_refspecs(
         self,
         url: str,
-        refspecs: Union[str, Iterable[str]],
+        refspecs: Union[bytes, List[bytes], List[Tuple[bytes, bytes]]],
```
pmrowla, line 244: In scmrepo our methods should take strings, not bytes, and do any conversion to bytes in the dulwich backend.

Basically the existing typehint for this method is what we want, and the dulwich implementation should be doing the conversion to
```
[os.fsencode(refspec) for refspec in refspecs]
```
when we call the underlying dulwich method

pmrowla (changes requested):

scmrepo/git/backend/base.py
pmrowla: also just to be clear, the fix for this also needs to account for when the user passes scmrepo a single string

pmrowla (approved):

PR #69, opened 2022-05-18
git: add untracked_files kwarg to status

dtrifiro:
note that `untracked_files` is only currently available in `dulwich>=0.20.38`

### DIFF
--- a/tests/test_git.py
+++ b/tests/test_git.py
@@ -988,3 +988,56 @@ def test_fetch(
 
     target.fetch()
     assert target.get_ref("refs/remotes/origin/master") == rev
+
+
+@pytest.mark.skip_git_backend("pygit2", "gitpython")
+@pytest.mark.parametrize("untracked_files", ["all", "no"])
+@pytest.mark.parametrize("ignored", [False, True])
+def test_status(
+    tmp_dir: TmpDir,
+    scm: Git,
+    git: Git,
+    tmp_dir_factory: TempDirFactory,
+    untracked_files: str,
+    ignored: bool,
+):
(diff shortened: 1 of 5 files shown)


tests/test_git.py
```diff
@@ -945,3 +988,64 @@ def test_fetch(
 
     target.fetch()
     assert target.get_ref("refs/remotes/origin/master") == rev
+
+
+@pytest.mark.skip_git_backend("pygit2", "gitpython")
+@pytest.mark.parametrize("untracked_files", ["all", "no"])
```
dtrifiro, line 994: I felt a bit conflicted when this test, because I feel like this is testing something that is already tested in `dulwich`, but this is basically testing that the `untracked_files` and `ignored` is correctly passed down to `dulwich`'s `status`

pmrowla: I'm not sure I follow?

Testing this here is correct because in scmrepo we should be testing that scmrepo's returned `status()` is the same regardless of our git backend.

tests/test_git.py
```diff
@@ -945,3 +988,56 @@ def test_fetch(
...
+    staged, unstaged, untracked = git.status(ignored, untracked_files)
+
+    assert not staged
+    assert not unstaged
+    if ignored and untracked_files != "no":
+        assert untracked == ["ignored"]
+    else:
+        assert not untracked
+
+    with (tmp_dir / "foo").open("a") as fh:
```
pmrowla, line 1023: Would prefer using `fobj` over `fh` so that it's consistent with the rest of the dvc codebase

pmrowla (approved):

PR #72, opened 2022-05-19
bump to 0.0.23

dtrifiro:


dtrifiro:
Actually, please hold on, there is one extra thing I want to do before bumping version :pray:

dtrifiro:
I'd be happy to also include https://github.com/iterative/scmrepo/pull/74 in 0.0.23

efiop:
@dtrifiro We also need to put an upper bound on dulwich, so we have the wheels.

dtrifiro:
I'll test `0.20.39`, and wait for that to be [tagged](https://github.com/jelmer/dulwich/tags)

skshetry:
Closing this, now there's no need to manually bump the version. Just create the release from GitHub UI, and it'll automatically publish (thanks to `setuptools_scm`).

PR #74, opened 2022-05-19
dulwich: use untracked_files="no" in is_dirty

dtrifiro:
This will speed up calling `is_dirty` when `untracked_files=False`

related to #69

### DIFF
--- a/scmrepo/git/backend/dulwich/__init__.py
+++ b/scmrepo/git/backend/dulwich/__init__.py
@@ -7,6 +7,7 @@ from functools import partial
 from io import BytesIO, StringIO
 from typing import (
     TYPE_CHECKING,
+    Any,
     Callable,
     Dict,
     Iterable,
@@ -344,8 +345,10 @@ class DulwichBackend(BaseGitBackend):  # pylint:disable=abstract-method
         return False
 
     def is_dirty(self, untracked_files: bool = False) -> bool:
-        staged, unstaged, untracked = self.status()
-        return bool(staged or unstaged or (untracked_files and untracked))
+        kwargs: Dict[str, Any] = (
+            {} if untracked_files else {"untracked_files": "no"}
+        )
+        return any(self.status(**kwargs))
 
     def active_branch(self) -> str:
         raise NotImplementedError
--- a/tests/test_git.py
+++ b/tests/test_git.py
@@ -1041,3 +1041,70 @@ def test_status(
     assert staged["modify"] == ["foo"]
     assert unstaged == ["bar"]
     assert untracked == expected_untracked
+
+
+@pytest.mark.skip_git_backend("pygit2")
+def test_is_dirty_empty(
+    tmp_dir: TmpDir,
+    scm: Git,
+    git: Git,
+):
+    assert not git.is_dirty()
+
+
+@pytest.mark.skip_git_backend("pygit2")
+def test_is_dirty_added(
+    tmp_dir: TmpDir,
+    scm: Git,
+    git: Git,
+):
+    tmp_dir.gen("foo", "foo")
+    scm.add("foo")
+
+    assert git.is_dirty()
+    scm.commit("commit")
+    assert not git.is_dirty()
+
+
+@pytest.mark.skip_git_backend("pygit2")
+def test_is_dirty_modified(
+    tmp_dir: TmpDir,
+    scm: Git,
+    git: Git,
+):
+    tmp_dir.gen("foo", "foo")
+    scm.add_commit("foo", message="add foo")
+
+    tmp_dir.gen("foo", "modified")
+    assert git.is_dirty()
+    scm.add_commit("foo", message="modified")
+    assert not git.is_dirty()
+
+
+@pytest.mark.skip_git_backend("pygit2")
+def test_is_dirty_deleted(
+    tmp_dir: TmpDir,
+    scm: Git,
+    git: Git,
+):
+    tmp_dir.gen("foo", "foo")
+    scm.add_commit("foo", message="add foo")
+
+    assert not git.is_dirty()
+    os.unlink(tmp_dir / "foo")
+    assert git.is_dirty()
+    scm.add("foo")
+    assert git.is_dirty()
+    scm.commit("deleted")
+    assert not git.is_dirty()
+
+
+@pytest.mark.skip_git_backend("pygit2")
+def test_is_dirty_untracked(
+    tmp_dir: TmpDir,
+    scm: Git,
+    git: Git,
+):
+    tmp_dir.gen("untracked", "untracked")
+    assert git.is_dirty(untracked_files=True)
+    assert not git.is_dirty(untracked_files=False)


efiop:
Was this tested with dvc?

dtrifiro:
> Was this tested with dvc?

Not yet, why?

efiop:
@dtrifiro Well, it is not good if it will break dvc tests 😄 Let's not release this until it is tested.

dtrifiro:
> @dtrifiro Well, it is not good if it will break dvc tests smile Let's not release this until it is tested.

:sweat_smile: indeed, although it does not change existing behaviour. Running tests right now

dtrifiro:
No tests fail when running locally: Works on my machine® 

Currently running some CI tests here with [`scmrepo@main`](https://github.com/dtrifiro/dvc/commit/4a06706a1a1d311a35be76cc594460a12a66de50):
https://github.com/dtrifiro/dvc/actions/runs/2352088728

scmrepo/git/backend/dulwich/__init__.py
```diff
@@ -344,8 +344,12 @@ def is_tracked(self, path: str) -> bool:
...
 
     def is_dirty(self, untracked_files: bool = False) -> bool:
-        staged, unstaged, untracked = self.status()
-        return bool(staged or unstaged or (untracked_files and untracked))
+        if untracked_files:
+            staged, unstaged, untracked = self.status()
+            return bool(staged or unstaged or untracked)
+
+        staged, unstaged, _ = self.status(untracked_files="no")
+        return bool(staged or unstaged)
```
efiop, line 352: We don't really promise backward compatibility at this point, so maybe consider adjusting the `is_dirty` flag

setup.cfg
```diff
@@ -25,7 +25,7 @@ zip_safe = False
 packages = find:
 install_requires=
     gitpython>3
-    dulwich>=0.20.38
+    dulwich<=0.20.39
```
efiop, line 28: let's add a note that dulwich build process is breaking from time to time and so to ensure that we will always have wheels, we will bump this manually from time to time

scmrepo/git/backend/dulwich/__init__.py
```diff
@@ -344,8 +344,9 @@ def is_tracked(self, path: str) -> bool:
         return False
 
     def is_dirty(self, untracked_files: bool = False) -> bool:
-        staged, unstaged, untracked = self.status()
-        return bool(staged or unstaged or (untracked_files and untracked))
+        kwargs = {} if untracked_files else {"untracked_files": "no"}
+        staged, unstaged, untracked = self.status(**kwargs)  # type: ignore
```
dtrifiro, line 348: Bit confused by mypy here. Found this: https://github.com/python/mypy/issues/11583

Can be replicated with this:

```python
def f(b: bool = True, **kwargs: str) -> None:
    pass

f(**{'x': 'x'})```

File: setup.cfg (hunk 2 above)
Comment (dtrifiro) [reply setup.cfg]: Status for `0.20.240`: https://github.com/jelmer/dulwich/actions/runs/2353092675

File: scmrepo/git/backend/dulwich/__init__.py (hunk 3 above)
Comment (pmrowla) [reply scmrepo/git/backend/dulwich/__init__.py:348]: I think mypy should pass if you explicitly type kwargs as 
```
kwargs: Dict[str, Any] = {} if ...
```

File: setup.cfg [hunk 4]
```diff
@@ -25,7 +25,7 @@ zip_safe = False
 packages = find:
 install_requires=
     gitpython>3
-    dulwich>=0.20.38
+    dulwich<=0.20.40 # set upper bound for dulwich so that we can check that wheels are available before bumping version
```
efiop, line 28:
```suggestion
    dulwich<=0.20.38
```
dulwich wheels are now available

tests/test_git.py
```diff
@@ -1041,3 +1041,50 @@ def test_status(
...
+
+    with (tmp_dir / "untracked").open("w") as fobj:
+        fobj.write("untracked")
+
+    backend = list(git.backends)[0]
+    assert backend in ("dulwich", "gitpython")
+    if backend == "dulwich":
+        spy = mocker.spy(git.dulwich, "status")
+    elif backend == "gitpython":
+        spy = mocker.spy(git.gitpython.repo, "is_dirty")
```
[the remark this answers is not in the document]
pmrowla: I don't think we need to be `spy`ing individual backend calls at all here. `test_git` tests should just be asserting whether or the relevant call (`is_dirty()` here) returns the expected results across each backend.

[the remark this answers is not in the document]
dtrifiro: Yeah, this is pretty ugly. I'd probably rather go to having one test per backend. There will be some setup-related duplicated code, but the tests themselves will be less confusing.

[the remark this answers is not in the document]
pmrowla: @dtrifiro Having two separate tests in `test_git.py` makes it look like we expect different results for `is_dirty()` depending on which backend we are using. IMO `test_git` should have one test that verifies that `is_dirty()` returns the correct values regardless of backend (and it should not test any backend-specific behavior).

I'm fine with adding a separate test for verifying the dulwich kwargs in `test_dulwich.py`, but I don't think that the gitpython specific test behavior adds anything beneficial (since gitpython's native `is_dirty` doesn't have the distinct options for different levels of untracked files behavior)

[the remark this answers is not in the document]
dtrifiro: I ended up scrapping the `spy`ing to keep things simpler, as the value of that check was questionable

scmrepo/git/backend/dulwich/__init__.py
```diff
@@ -344,8 +345,11 @@ def is_tracked(self, path: str) -> bool:
         return False
 
     def is_dirty(self, untracked_files: bool = False) -> bool:
-        staged, unstaged, untracked = self.status()
-        return bool(staged or unstaged or (untracked_files and untracked))
+        kwargs: Dict[str, Any] = (
+            {} if untracked_files else {"untracked_files": "no"}
+        )
+        staged, unstaged, untracked = self.status(**kwargs)
+        return bool(staged or unstaged or untracked)
```
skshetry, line 352: `any` makes the intention more clearer.
```suggestion
        staged, unstaged, untracked = self.status(**kwargs)
        return any((staged, unstaged, untracked))
```

You can further simplify this to just:

```suggestion
        return any(self.status(**kwargs))
```

or, 
```

File: tests/test_git.py [hunk 7]
```diff
@@ -1041,3 +1041,32 @@ def test_status(
...
+        {
+            "foo": "foo",
+        }
+    )
+    scm.add_commit(["foo"], message="init")
+
+    assert not git.is_dirty()
+
+    with (tmp_dir / "foo").open("a") as fobj:
+        fobj.write("modified")
```
skshetry, line 1064: For consistency,
```suggestion
    tmp_dir.gen("foo", "modified")
```

tests/test_git.py
```diff
@@ -1041,3 +1041,32 @@ def test_status(
     assert staged["modify"] == ["foo"]
     assert unstaged == ["bar"]
     assert untracked == expected_untracked
+
+
+@pytest.mark.skip_git_backend("pygit2")
+@pytest.mark.parametrize("untracked_files", [False, True])
+def test_is_dirty(
```
skshetry, line 1048: I'll suggest splitting this test into 2 tests:
1. test for untracked
2. test for modify-then-commit.

dtrifiro: Ended up getting rid of the parametrization and adding more tests for `is_dirty`

skshetry: Please try to split these blocks into individual tests if possible.

scmrepo/git/backend/dulwich/__init__.py
```diff
@@ -344,8 +345,10 @@ def is_tracked(self, path: str) -> bool:
         return False
 
     def is_dirty(self, untracked_files: bool = False) -> bool:
-        staged, unstaged, untracked = self.status()
-        return bool(staged or unstaged or (untracked_files and untracked))
+        kwargs: Dict[str, Any] = (
+            {} if untracked_files else {"untracked_files": "no"}
+        )
+        return any(self.status(**kwargs))
```
skshetry, line 351: I wonder if we should optimize this further, i.e. use generator and return False if there is any changes rather than collecting everything. :)

Something like:
```python
from itertools import chain, zip_longest

with open_repo_closing(repo) as r:
        # 1. Get status of staged
        tracked_changes = get_tree_changes(r)
        # 2. Get status of unstaged
        index = r.open_index()
        normalizer = r.get_blob_normalizer()
        filter_callback = normalizer.checkin_normalize

        unstaged_changes = get_unstaged_changes(index, r.path, filter_callback)
        untracked_paths = get_untracked_paths(
            r.path,
            r.path,
            index,
            exclude_ignored=not ignored,
            untracked_files=untracked_files,
        )
        return any(chain.from_iterable(zip_longest(s, d)))
```

The only doubt that I have is that it won't reuse IgnoreManager, which I think already happens for `status`.
No strong opinion though, we could also propose this in dulwich.

skshetry (approved):
LGTM 🙂

efiop:
@pmrowla Could you take a final look as well, please?

PR #82, opened 2022-05-31
objects: don't recompute object size

efiop:
Was discovered as a part of https://github.com/iterative/dvc/issues/7824

### DIFF
--- a/scmrepo/git/backend/pygit2.py
+++ b/scmrepo/git/backend/pygit2.py
@@ -62,8 +62,8 @@ class Pygit2Object(GitObject):
             return stat.S_IFDIR
         return self.obj.filemode
 
-    @property
-    def size(self) -> int:
+    @cached_property
+    def size(self) -> int:  # pylint: disable=invalid-overridden-method
         try:
             return len(self.obj.read_raw())
         except KeyError:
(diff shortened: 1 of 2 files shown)


scmrepo/git/backend/pygit2.py
```diff
@@ -3,6 +3,7 @@
 import os
 import stat
 from contextlib import contextmanager
+from functools import cached_property
```
skshetry, line 6: I'll suggest sticking to `funcy.cached_property` instead due to https://github.com/python/cpython/issues/87634.

Unless we are trying to replace the following piece of code:
```python
@funcy.wrap_prop(threading.RLock())
@funcy.cached_property

# equivalent to @functools.cached_property
```

efiop: Makes sense, was just trying it out with pylint and 3.8

efiop: Didn't know about that bug though 🙁 Thank you!

scmrepo/git/backend/pygit2.py
```diff
@@ -62,8 +62,8 @@ def mode(self):
             return stat.S_IFDIR
         return self.obj.filemode
 
-    @property
-    def size(self) -> int:
+    @cached_property
+    def size(self) -> int:  # pylint: disable=invalid-overridden-method
         try:
             return len(self.obj.read_raw())
```
efiop, line 68: Welp, actually there is size, but just not for trees. Will send a followup.

PR #84, opened 2022-05-31
dulwich: fix DulwichObject

efiop:
Will make the test suit run test_fs for multiple backends in a separate PR.


scmrepo/git/backend/dulwich/__init__.py
```diff
@@ -72,19 +72,22 @@ def mode(self) -> int:
...
-        return len(self.repo[self.sha].as_raw_string())
+        try:
+            return self.repo[self._sha].raw_length()
+        except KeyError:
+            return 0
 
-    @property
-    def sha(self) -> str:
-        return self._sha
+    @cached_property
```
skshetry, line 88: Do we really need a `cached_property` here?

skshetry: @efiop, could you please share some benchmarks?

efiop: @skshetry No benchmarks there. Dulwich is not even used usually here (pygit2 is). Just from a correctness perspective.

PR #85, opened 2022-05-31
tests: fs: test all backends

efiop:


scmrepo/fs.py
```diff
@@ -172,7 +172,7 @@ def __init__(
             scm = scm or Git(path)
             resolver = rev_resolver or Git.resolve_rev
             resolved = resolver(scm, rev or "HEAD")
-            tree_obj = scm.pygit2.get_tree_obj(rev=resolved)
```
efiop, line 175: Hm, looks like we were always force-using pygit2 during our normal operation. This might have an unexpected impact...

scmrepo/git/__init__.py
```diff
@@ -363,6 +363,8 @@ def add_commit(
     validate_git_remote = partialmethod(_backend_func, "validate_git_remote")
     check_ref_format = partialmethod(_backend_func, "check_ref_format")
 
+    get_tree_obj = partialmethod(_backend_func, "get_tree_obj")
```
efiop, line 366: This is only used in tests now, but in the future we should make `get_fs` not default to `GitFileSystem`'s default behaviour that uses pygit2.

efiop:
Note https://github.com/iterative/scmrepo/issues/86

efiop:
Guys, I'll merge in the interest of time, but would appreciate if you still take a look. Thank you 🙏

PR #87, opened 2022-05-31
fs: compute size lazily

efiop:
We are effectively reading every object in the tree that we parse right now,
which takes a very long time.

Before fsspec migration, we used to not use info/size during walking at all.

Cuts down `dvc exp show -A` time for me from ~50sec down to ~3sec.

Related iterative/dvc#7824

### DIFF
--- a/scmrepo/git/objects.py
+++ b/scmrepo/git/objects.py
@@ -1,7 +1,7 @@
 import stat
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import Iterable, List, Optional
+from typing import Iterable, List, Optional, cast
 
 from pygtrie import Trie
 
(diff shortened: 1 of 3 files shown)


scmrepo/git/objects.py
```diff
@@ -136,13 +136,21 @@ def walk(self, top: tuple, topdown: Optional[bool] = True):
...
-            "size": obj.size,
-            "type": "directory" if stat.S_ISDIR(obj.mode) else "file",
-            "sha": obj.sha,
-            "mode": obj.mode,
-        }
+
+        def size():
+            return obj.size
+
+        return LazyDict(
```
efiop, line 146: Obviously, we could create a flag for `info` and then override `fs.walk`(and some others) to use it, but I wanted to try out this lazy approach, to see if it will even work in this setup. Looks like it does, but I might go back to a more explicit approach anyways (will get back to this in the morning).

efiop: Using a flag is somewhat too explicit and requires overriding a lot of methods. Let's try the `LazyDict` for now and see how that goes.

PR #88, opened 2022-06-13
gha: run tests in pypy

skshetry:


skshetry:
Tests are failing due to https://foss.heptapod.net/pypy/pypy/-/issues/3168.

efiop:
@skshetry looks like just need to implicitly covert in our tests and that should fix it?

skshetry:
yes, we need to convert pathlib to fspaths ourselves.

skshetry:
> @skshetry looks like just need to implicitly covert in our tests and that should fix it?

This is something that needs to be fixed in pygit2; we need to add compat APIs for supporting Path objects in PyPy.

PR #101, opened 2022-06-17
stash apply: support reinstate_index and skip_conflicts

pmrowla:
Adds support for `--index` and force-applying in `stash apply`

prereq for https://github.com/iterative/dvc/issues/6930

### DIFF
--- a/scmrepo/git/backend/pygit2.py
+++ b/scmrepo/git/backend/pygit2.py
@@ -465,16 +465,25 @@ class Pygit2Backend(BaseGitBackend):  # pylint:disable=abstract-method
             self.repo.stash_drop()
         return str(oid), False
 
-    def _stash_apply(self, rev: str):
-        from pygit2 import GitError
+    def _stash_apply(
+        self,
+        rev: str,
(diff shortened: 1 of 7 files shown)


scmrepo/git/backend/pygit2.py
```diff
@@ -465,16 +465,25 @@ def _stash_push(
...
+        from pygit2 import GIT_CHECKOUT_ALLOW_CONFLICTS, GitError
 
         from scmrepo.git import Stash
 
         def _apply(index):
             try:
                 self.repo.index.read(False)
+                strategy = self._get_checkout_strategy()
+                if skip_conflicts:
+                    strategy |= GIT_CHECKOUT_ALLOW_CONFLICTS
```
pmrowla, line 484: Note that this is pygit specific right now because there is no CGit `git stash apply` flag that gives this equivalent behavior.

libgit2 provides it since it's useful for programmatic reasons (where you need to automate everything vs being able to do interactive 3-way merge conflict resolutions with CLI git)

PR #118, opened 2022-07-19
pygit2: implement status

dtrifiro:
- bump pygit2 to 1.10.0
- force dulwich.status to return posix-style paths for untracked files

### DIFF
--- a/scmrepo/git/backend/pygit2.py
+++ b/scmrepo/git/backend/pygit2.py
@@ -34,11 +34,6 @@ logger = logging.getLogger(__name__)
 if TYPE_CHECKING:
     from scmrepo.progress import GitProgressEvent
 
-# NOTE: constant from libgit2 git2/checkout.h
-# This can be removed after next pygit2 release:
-# see https://github.com/libgit2/pygit2/pull/1087
-GIT_CHECKOUT_SKIP_LOCKED_DIRECTORIES = 1 << 18
-
 
 class Pygit2Object(GitObject):
     def __init__(self, obj):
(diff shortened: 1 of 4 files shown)


dtrifiro:
Please note that tests in windows are failing due to dulwich returning nt-style paths for untracked files on windows instead of returning posix-style paths (like cli git does).

skshetry:
[`pygit2 v1.10.0`](https://github.com/libgit2/pygit2/releases/tag/v1.10.0) is now released.

tests/test_git.py
```diff
@@ -939,17 +939,26 @@ def test_fetch(
     assert target.get_ref("refs/remotes/origin/master") == rev
 
 
-@pytest.mark.skip_git_backend("pygit2", "gitpython")
-@pytest.mark.parametrize("untracked_files", ["all", "no"])
+@pytest.mark.skip_git_backend("gitpython")
+@pytest.mark.parametrize("untracked_files", ["all", "no", "normal"])
 @pytest.mark.parametrize("ignored", [False, True])
 def test_status(
+    request,
```
skshetry, line 946: You can get currently used backend from `git_backend` fixture.

skshetry: I see that you are using that but indirectly with `getfixturevalue`.

scmrepo/git/backend/pygit2.py
```diff
@@ -34,11 +34,6 @@
 if TYPE_CHECKING:
     from scmrepo.progress import GitProgressEvent
 
-# NOTE: constant from libgit2 git2/checkout.h
-# This can be removed after next pygit2 release:
-# see https://github.com/libgit2/pygit2/pull/1087
-GIT_CHECKOUT_SKIP_LOCKED_DIRECTORIES = 1 << 18
-
```
dtrifiro, line 41: Just some housekeeping

tests/test_git.py
```diff
@@ -939,17 +939,24 @@ def test_fetch(
...
```
skshetry, line 946: This is unused.

dtrifiro:
While fixing this, I noticed something weird: since `reset` is not implemented for dulwich, in `estst:
https://github.com/iterative/scmrepo/blob/f35b124b85125665084928a0bb22c2fd1841d3f4/tests/test_git.py#L659-L681

The `reset` call switches backend from `dulwich` to `pygit2`, and since this PR adds `pygit2.status`, the `scm.status` call uses dulwich while the second call uses `pygit2`. I guess this is one of these issues that only come up when switching backends.

skshetry:
@dtrifiro, in the future, please don't create a PR from your default branch. :)

skshetry (approved):

Issue #144, opened 2022-10-19
Add multi rev support for function `describe` 


karajan1001:
In our current `describe` function we only accept one rev at a time and will try to get a reference for every reference in the repo. It will cost O(MN) in `exp show` in which we try to get a ref for every single experiment. 

https://github.com/iterative/scmrepo/blob/ed1798525392f9850082ddf2ad3fb01897660ccc/src/scmrepo/git/backend/dulwich/__init__.py#L725-L740

![image](https://user-images.githubusercontent.com/6745454/196609937-0e5f5bcb-a2e3-409d-a389-9d5db235f6ca.png)

PR #145, opened 2022-10-19
Add mutli rev support for function `describe`

karajan1001:
fix: #144

In our current describe function we only accept one rev at a time and will try to get a reference for every reference in the repo. It will cost O(MN) in the `exp show` in which we try to get a ref for every single experiment.

### DIFF
--- a/src/scmrepo/git/__init__.py
+++ b/src/scmrepo/git/__init__.py
@@ -442,15 +442,23 @@ class Git(Base):
 
     def describe(
         self,
-        rev: str,
+        revs: Iterable[str],
         base: Optional[str] = None,
         match: Optional[str] = None,
         exclude: Optional[str] = None,
-    ) -> Optional[str]:
-        if (
-            base == "refs/heads"
-            and self.get_rev() == rev
-            and self.get_ref("HEAD", follow=False).startswith(base)
-        ):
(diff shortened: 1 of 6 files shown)


skshetry:
@karajan1001, can you please post the new results of the profiling with this patch in the same way that you did on #144?

Also before/after runtime for a real `dvc exp show` would be great.

karajan1001:
> @karajan1001, can you please post the new results of the profiling with this patch in the same way that you did on #144?
> 
> Also before/after runtime for a real `dvc exp show` would be great.

In https://github.com/iterative/dvc/pull/8453.  In the same repo, this version of describe runs 5 times faster than the previous one. (With the experiment) grows the difference will be even bigger.

skshetry (commented):
I don't think I have enough understanding about this on the `exp` side. So I'll leave this up to @pmrowla. 

I will suggest considering `RefsIterator` or something similar concept that allows us to do this kind of thing, as it seems this way we may need to extend multiple APIs to support lists, breaking the API. Also `repo.refs` of `dulwich` comes to mind, but it's limited (but we can extend :) ).

Anyway, I'm okay with this for the time being. :)

pmrowla (approved):

Issue #154, opened 2022-11-28
Wrap `Keyrror` from `fetch_refspec` into `SCMError`. 


karajan1001:
_Originally posted by @karajan1001 in https://github.com/iterative/dvc/issues/8615#issuecomment-1328586415_

PR #155, opened 2022-11-28
Wrap  KeyError from fetch_refspec into SCMError.

karajan1001:
fix: #154

1. Wrap KeyError from fetch_refspec into SCMError.
2. Modify the test to test this change.


pmrowla (approved):

karajan1001:
Looks like also suffer from the https://github.com/iterative/py-template/pull/92

PR #159, opened 2022-12-15
improve auth error on clone

dtrifiro:
- exceptions: update AuthError message
- dulwich: add AuthError on HTTPUnauthorized

fixes iterative/dvc#8654

### DIFF
--- a/src/scmrepo/git/backend/dulwich/__init__.py
+++ b/src/scmrepo/git/backend/dulwich/__init__.py
@@ -531,8 +531,8 @@ class DulwichBackend(BaseGitBackend):  # pylint:disable=abstract-method
                 yield from (os.fsdecode(ref) for ref in client.get_refs(path))
         except NotGitRepository as exc:
             raise InvalidRemote(url) from exc
-        except HTTPUnauthorized:
-            raise AuthError(url)
+        except HTTPUnauthorized as exc:
+            raise AuthError(url) from exc
 
     def get_refs_containing(self, rev: str, pattern: Optional[str] = None):
         raise NotImplementedError
@@ -616,8 +616,8 @@ class DulwichBackend(BaseGitBackend):  # pylint:disable=abstract-method
         except (NotGitRepository, SendPackError) as exc:
             src = [lh for (lh, _, _) in selected_refs]
             raise SCMError(f"Git failed to push '{src}' to '{url}'") from exc
-        except HTTPUnauthorized:
-            raise AuthError(url)
+        except HTTPUnauthorized as exc:
+            raise AuthError(url) from exc
         return change_result
 
     def fetch_refspecs(
--- a/src/scmrepo/exceptions.py
+++ b/src/scmrepo/exceptions.py
@@ -27,7 +27,7 @@ class InvalidRemote(SCMError):
 class AuthError(SCMError):
     def __init__(self, url: str) -> None:
         self.url = url
-        super().__init__(f"HTTP Git authentication is not supported: '{url}'")
+        super().__init__(f"Authentication failed for: '{url}'")
 
 
 class CloneError(SCMError):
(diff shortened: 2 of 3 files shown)


src/scmrepo/git/backend/dulwich/__init__.py
```diff
@@ -199,6 +200,8 @@ def clone(
             else:
                 repo = clone_from()
             cls._set_default_tracking_branch(repo)
+        except HTTPUnauthorized:
+            raise AuthError(url)
```
skshetry, line 204:
```suggestion
        except HTTPUnauthorized as exc:
            raise AuthError(url) from exc
```

dtrifiro: 👍 Also changed other instances of AuthError

dtrifiro:
> Minor: Can we also include the exception name (dulwich.client.HTTPUnauthorized)? I think it's helpful to see where it originated. It can be hard to debug because we are wrapping other exceptions with our own exceptions.

`raise AuthError from exc` will show the dulwich exception in the traceback (although only when used with `-v`).

I think the `AuthError` with the `Authentication failed` message makes it a bit clearer, and extra information can always be retrieved with `-v`.

If you think the dulwich exception is clearer, we could also just stop catching `HTTPUnauthorized` so that it is re-raised as it is (in `clone`, `iter_remote_refs` and `push_refspecs`, which are related to experiments)

dberenbaum:
Thanks @dtrifiro!

> If you think the dulwich exception is clearer, we could also just stop catching `HTTPUnauthorized` so that it is re-raised as it is (in `clone`, `iter_remote_refs` and `push_refspecs`, which are related to experiments)

I think it's clearer to show `Failed to clone repo...` in addition to some authentication error. I'm not sure it's obvious to all users that dvc is doing a git clone, so I'd rather raise the `CloneError` along with the info from the original exception.

More importantly, I'd rather do this for other exceptions so that any other clone failures can be more easily debugged. I know it's maybe duplicating the point of `-v`, but I think it's generally useful enough to be worth including and saving people from needing `-v` in some cases.

For example, instead of:

```
ERROR: failed to get 'vimrc' from 'git@github.com:..' - Failed to clone repo 'git@github.com:...' to '/var/folders/...'
```

It would be more helpful to immediately see:

```
ERROR: failed to get 'vimrc' from 'git@github.com:..'
Failed to clone repo 'git@github.com:...' to '/var/folders/...'
Caused by: FileNotFoundError: [Errno 2] No such file or directory: '/Users/dave/.ssh/id_ed25519'
```

One more minor thing - the `AuthError` still shows as an 'unexpected error', unlike the `CloneError`, although I don't know why.

dtrifiro:
How about something like this:
>ERROR: failed to get 'README.md' from 'https://github.com/dberenbaum/example-get-started.git' - Failed to clone repo 'https://github.com/dberenbaum/example-get-started-cp.git' to '/tmp/tmpmxbwwvwmdvc-clone' - No valid credentials provided

dberenbaum:
> How about something like this:

The error message looks good, thanks!

> More importantly, I'd rather do this for other exceptions so that any other clone failures can be more easily debugged.

What do you think about this?

skshetry:
dvc is responsible for error messages, not the scmrepo.

src/scmrepo/git/backend/dulwich/__init__.py
```diff
@@ -199,6 +200,8 @@ def clone(
             else:
                 repo = clone_from()
             cls._set_default_tracking_branch(repo)
+        except HTTPUnauthorized as exc:
+            raise AuthError(url) from exc
```
skshetry, line 204: For the clone, since we already have `CloneError`, it might be better to reuse that and only change the exception message.

skshetry:
> One more minor thing - the `AuthError` still shows as an 'unexpected error', unlike the `CloneError`, although I don't know why.

`CloneError` is handled in dvc and reraised as dvc's error.

src/scmrepo/git/backend/dulwich/__init__.py
dtrifiro: Yep, that's what I was thinking

dberenbaum (approved):

PR #169, opened 2023-01-17
implement pygit2 backend for fetch_refspec

### LINKED ISSUE #168: Implement pygit2 backends for fetch_refspec
In https://github.com/iterative/dvc/issues/8477 the Dulwich backend of `fetch_refspec` is the bottleneck of `dvc exp show`.  Implementing a `pygit2` version of `fetch_refspec` can solve this problem.

karajan1001:
implement pygit2 backend for fetch_refspec

fix #168 

Compare to the `cprofile` in https://github.com/iterative/dvc/issues/8477. 

The new greatly reduced the time cost in fetching.

image

Making the total `dvc exp show` query finished in 1s. And brings a much smoother performance in vs-code extension.  And also solves the hanging problem in https://github.com/iterative/dvc/issues/8762

https://user-images.githubusercontent.com/6745454/212892100-ae2d032d-e23b-4b62-9207-28752a3de4c8.mp4

But we still had one problem remaining to solve, how to make `scmrepo` use the `pygit2` as the default backend. In the test and the demo, I simply raise `NotImplementError` in Dulwich to disable it.

### DIFF
--- a/src/scmrepo/git/__init__.py
+++ b/src/scmrepo/git/__init__.py
@@ -41,6 +41,8 @@ class GitBackends(Mapping):
         """Lazily initialize backends and cache it afterwards"""
         initialized = self.initialized.get(key)
         if not initialized:
+            if key not in self.backends and key in self.DEFAULT:
+                raise NotImplementedError
             backend = self.backends[key]
             initialized = backend(*self.args, **self.kwargs)
             self.initialized[key] = initialized
@@ -266,11 +268,13 @@ class Git(Base):
     # https://github.com/iterative/dvc/issues/5641
     # https://github.com/iterative/dvc/issues/7458
     def _backend_func(self, name, *args, **kwargs):
-        for key, backend in self.backends.items():
+        backends: Iterable[str] = kwargs.pop("backends", self.backends)
+        for key in backends:
             if self._last_backend is not None and key != self._last_backend:
                 self.backends[self._last_backend].close()
                 self._last_backend = None
             try:
+                backend = self.backends[key]
                 func = getattr(backend, name)
                 result = func(*args, **kwargs)
                 self._last_backend = key
@@ -333,7 +337,9 @@ class Git(Base):
     iter_remote_refs = partialmethod(_backend_func, "iter_remote_refs")
     get_refs_containing = partialmethod(_backend_func, "get_refs_containing")
     push_refspecs = partialmethod(_backend_func, "push_refspecs")
-    fetch_refspecs = partialmethod(_backend_func, "fetch_refspecs")
+    fetch_refspecs = partialmethod(
+        _backend_func, "fetch_refspecs", backends=["pygit2", "dulwich"]
+    )
     _stash_iter = partialmethod(_backend_func, "_stash_iter")
     _stash_push = partialmethod(_backend_func, "_stash_push")
     _stash_apply = partialmethod(_backend_func, "_stash_apply")
--- a/src/scmrepo/git/backend/pygit2.py
+++ b/src/scmrepo/git/backend/pygit2.py
@@ -7,6 +7,8 @@ from io import BytesIO, StringIO
 from typing import (
     TYPE_CHECKING,
     Callable,
+    Dict,
+    Generator,
     Iterable,
     List,
     Mapping,
@@ -15,7 +17,8 @@ from typing import (
     Union,
 )
 
-from funcy import cached_property
+from funcy import cached_property, reraise
+from shortuuid import uuid
 
 from scmrepo.exceptions import CloneError, MergeConflictError, RevError, SCMError
 from scmrepo.utils import relpath
@@ -27,6 +30,8 @@ logger = logging.getLogger(__name__)
 
 
 if TYPE_CHECKING:
+    from pygit2.remote import Remote  # type: ignore
+
     from scmrepo.progress import GitProgressEvent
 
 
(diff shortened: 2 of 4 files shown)


pmrowla:
> But we still had one problem remaining to solve, how to make `scmrepo` use the `pygit2` as the default backend. In the test and the demo, I simply raise `NotImplementError` in Dulwich to disable it.

Setting the default backend is controlled here:
https://github.com/iterative/scmrepo/blob/339fdba37383888cd4e5d20dc729efdb509cd919/src/scmrepo/git/__init__.py#L39-L41

But what we probably need is a way to just force preferring a specific backend in our `_backend_func` wrapper
https://github.com/iterative/scmrepo/blob/339fdba37383888cd4e5d20dc729efdb509cd919/src/scmrepo/git/__init__.py#L282

so we could do something like
```
scm.fetch_refspec(..., prefer_backend="pygit2")
```
which would force rotating the backends list until `pygit2` is at the top before continuing to try `fetch_refspec` calls

skshetry:
> 
> so we could do something like
> 
> ```python
> scm.fetch_refspec(..., prefer_backend="pygit2")
> ```

Instead of a wrapper, we could also use it directly as following in dvc:
```python
scm.pygit2.fetch_refspecs(...)
```

karajan1001:
> ```
> scm.fetch_refspec(..., prefer_backend="pygit2")
> ```

I had considered before, but the problem is that the backends will be determined the first time we call it. So if we want to make sure a function using a special backend, we need to add this prefer to all place we call it. So maybe it's better to add the argument in https://github.com/iterative/scmrepo/blob/339fdba37383888cd4e5d20dc729efdb509cd919/src/scmrepo/git/__init__.py#L282

dtrifiro:
Isn't using it directly also an issue with backend switching? We did have issues in the past related to that (https://github.com/iterative/dvc/issues/5641, https://github.com/iterative/dvc/issues/7458)

pmrowla:
Using the backends directly can cause issues due to caching of the git index, but fetch does not touch the git index at all so it shouldn't be a problem here

skshetry:
Does this work on Windows? AFAIK there is no ssh support in pygit2 for Windows?

src/scmrepo/git/backend/pygit2.py
```diff
@@ -459,7 +496,63 @@ def fetch_refspecs(
...
+        try:
+            remote: "Remote" = self.repo.remotes[url]
+        except ValueError:
+            remote_name = "fetch"
+            self.repo.remotes.set_url(remote_name, url)
+            remote = self.repo.remotes[remote_name]
+        except KeyError:
+            raise SCMError(f"'{url}' is not a valid Git remote or URL")
+
+        if os.name == "nt" and remote.url.startswith("file://"):
```
karajan1001, line 513: Not sure if should I put this code here or in the tests? Looks like `pygit2` doesn't support this format on the Windows platform.

pmrowla: it should support it https://github.com/libgit2/libgit2/blob/1327dbcf2a4273a8ba6fd978db5f0882530af94d/src/libgit2/transport.c#L35

pmrowla:
libgit2 supports ssh but only when it is compiled with libssh2. The libgit2 build bundled with the pygit2 wheels does not include libssh2, so `ssh://` is not supported in pygit2 https://github.com/libgit2/pygit2/issues/994

We should be checking to see if we are passed an ssh URL and then raise NotImpelementedError in pygit2 fetch_refspecs for ssh.

src/scmrepo/git/backend/pygit2.py
karajan1001: > it should support it https://github.com/libgit2/libgit2/blob/1327dbcf2a4273a8ba6fd978db5f0882530af94d/src/libgit2/transport.c#L35

But the test result is that on Mac or Linux it supports it, but not on Windows. I guess it's a bug of pygit2.

pmrowla: It might be that libgit2 or pygit2 expect windows slashes instead of posix ones after the `file://` protocol, or it might be that the `file://` prefix is supposed to be stripped as you are doing here. But in regular CLI git there is actually different behavior depending on whether you provide a `file://` URL or a raw local path (without `file://`), so we should double check what the problem is.

Please try to take a look into what pygit2 is doing in this situation

karajan1001: Hadn't found any issues related to it. Let me create a issue for them https://github.com/libgit2/pygit2/issues/1187

pmrowla: This could still be due to not creating a proper `Remote()` object for the raw URL. As we discussed, we should be using [`git_remote_create_anonymous`](https://libgit2.org/libgit2/#HEAD/group/remote/git_remote_create_anonymous) to generate our remote here.

```py
remote_name = "fetch"
self.repo.remotes.set_url(remote_name, url)
remote = self.repo.remotes[remote_name]
```

This is likely going to cause undefined behavior because we are trying to call `set_url` on a remote that doesn't actually exist (unless the user has already defined their own remote named `fetch`, in which case we are incorrectly modifying an existing git remote instead of creating our own temporary one)

karajan1001: Looks like `[git_remote_create_anonymous](https://libgit2.org/libgit2/#HEAD/group/remote/git_remote_create_anonymous)` are not implemented in `pygit2`, I just make a temporary solution.

pmrowla: Right, as we discussed though we should really consider adding it to `pygit`

src/scmrepo/git/backend/pygit2.py
```diff
@@ -412,6 +417,53 @@ def push_refspecs(
...
+
+    @contextmanager
+    def get_remote(self, url: str) -> Generator["Remote", None, None]:
+        try:
+            yield self.repo.remotes[url]
+        except ValueError:
+            remote_name = uuid()
+            self.repo.remotes.set_url(remote_name, url)
+            try:
+                yield self.repo.remotes[remote_name]
```
pmrowla, line 461: This still has the same issue where you will call `set_url` on a remote that doesn't actually exist which will likely end up causing errors and/or undefined behavior in the underlying libgit C library.

If you are going to implement it this way and create a full named remote (instead of adding the anonymous remote function call to pygit), you need to be using `self.repo.remotes.create()` here

src/scmrepo/git/__init__.py
```diff
@@ -266,17 +266,19 @@ def no_commits(self):
...
                 self._last_backend = None
             try:
+                backend = self.backends[key]
                 func = getattr(backend, name)
                 result = func(*args, **kwargs)
                 self._last_backend = key
                 self.backends.move_to_end(key, last=False)
                 return result
-            except NotImplementedError:
+            except (NotImplementedError, KeyError):
```
pmrowla, line 281: What `KeyError` is being hit here? If someone tries to call this function with `backends=[...]` that contains an invalid backend we should really be raising that exception and not catching/ignoring it here

karajan1001: It's related to the tests when we initialize the `backends` with `selected` argument, I had put the logic into `__getitem__` to raise different exceptions for the nonexistent backend names and the not initialized backend names.

src/scmrepo/git/backend/pygit2.py
```diff
@@ -421,7 +472,58 @@ def fetch_refspecs(
...
+            refspecs = [refspecs]
+
+        with self.get_remote(url) as remote:
+            if os.name == "nt" and remote.url.startswith("ssh://"):
+                raise NotImplementedError
+
+            if os.name == "nt" and remote.url.startswith("file://"):
+                url = remote.url[len("file://") :]
+                self.repo.remotes.set_url(remote.name, url)
+                remote = self.repo.remotes[remote.name]
```
pmrowla, line 487: Is this still needed even after the `remotes.create()` change?

karajan1001: Looks like it still fails.

pmrowla: You can make this change in the PR then, but we probably need to open another issue for it in scmrepo (this likely means that a regular git remote with `file://` URL in it is also broken for us on windows). This may still be related to us not setting the path separators or formatting the URL in a way that libgit2 expects and probably needs some more investigation.

pmrowla (approved):
@karajan1001 please open scmrepo issues for the pygit follow ups, we don't have to do them right away but it would be good for us to handle them whenever we have the bandwidth

- expose `git_remote_create_anonymous` in `pygit2.Remote.create_anonymous`
- investigate windows `file://` behavior in pygit/libgit

pmrowla:
@karajan1001 is this waiting on anything else or can it be merged?

karajan1001:
> @karajan1001 is this waiting on anything else or can it be merged?

I'll do some finally check today together with DVC.

pmrowla (approved):

Issue #176, opened 2023-02-01
Problems remained in `pygit2` implementation of `fetch_refspec` 


karajan1001:
@karajan1001 please open scmrepo issues for the pygit follow ups, we don't have to do them right away but it would be good for us to handle them whenever we have the bandwidth

- [ ]  expose `git_remote_create_anonymous` in `pygit2.Remote.create_anonymous`
- [ ] investigate windows `file://` behavior in pygit/libgit

_Originally posted by @pmrowla in https://github.com/iterative/scmrepo/pull/169#pullrequestreview-1278378606_

PR #177, opened 2023-02-06
Solve the url change problem in `pygit2` fetch_refspec

karajan1001:
Because for now `pygit2` fetch didn't support `file://` format on Windows, we removed this prefix. But didn't add it back after the fetch operation finished, and it makes the following `dulwich` backend goes wrong.

So here in the PR, we always create a new remote and delete it after the operation to make the remote config unchanged after the operation. Can be removed after #176 solved.

Related to the CI failure in https://github.com/iterative/dvc/actions/runs/4082287489/jobs/7037469526#logs


pmrowla:
This is probably another indicator that we should really be using `git_remote_add_anonymous` (since we technically are breaking the config until we actually hit our `finally:` cleanup block)

pmrowla (approved):

Issue #178, opened 2023-02-06
Credentials not parsed when git configuration contains unexpanded file paths


mgrbyte:
Credentials not parsed correctly when `credential.helper` is configured as the ` store` type  in a users' git-config, when pointing to a path that contains a `~` .e.g:

```text
...
[credential]
        helper = store --file ~/.git-credentials
....
```
The git repositories I'm working with are hosted on-prem w/gitlab, and must be accessed using credentials over https due toi nternal infrastructure reasons (thus using ssh keys is not an option).

The work-around is to use an absolute path in your git-config then the error above goes away.

I diagnosed by inserting `import pdb;pdb.set_trace()` [here](https://github.com/iterative/scmrepo/blob/a2ba90056c0022ff0166af916537dd352f12dec8/src/scmrepo/git/backend/dulwich/credentials.py#L179) and inspecting the result of the return value of the `subprocess.run` invocation.

> Which operating system and Python version are you using?

Both Python 3.10.6 (Host), Python 3.9.16 (Inside docker container) from [mambaforge](https://github.com/conda-forge/miniforge#mambaforge)

> Which version of this project are you using?

`0.1.7 (pip)`  (In both python environments described above)

> What did you do?

Ran `dvc update` and `dvc pull` against `.dvc` files containing `url` entries containing git urls using git `credentialshelper` configuration in place as described in this issue.

> What did you expect to see?

The commands complete without authentication errors due to missing credentials.

> What did you see instead?

This error (identical to that shown in "actual output" in as also reported [here ](https://github.com/jelmer/dulwich/issues/873#issue-902290509) when running various `dvc update` and `dvc pull` commands .

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/···/site-packages/dulwich/porcelain.py", line 1618, in ls_remote
    return client.get_refs(host_path)
  File "/···/site-packages/dulwich/client.py", line 2089, in get_refs
    refs, _, _ = self._discover_references(b"git-upload-pack", url)
  File "/···/site-packages/dulwich/client.py", line 1906, in _discover_references
    resp, read = self._http_request(url, headers, allow_compression=True)
  File "/···/site-packages/dulwich/client.py", line 1875, in _http_request
    raise HTTPUnauthorized(resp.getheader("WWW-Authenticate"), url)
dulwich.client.HTTPUnauthorized: No valid credentials provided
```

pmrowla:
[...]

PR #180, opened 2023-02-06
Fixes iterative/scmrepo#178.

mgrbyte:


src/scmrepo/git/backend/dulwich/credentials.py.passing
```diff
@@ -0,0 +1,267 @@
+# credentials.py -- support for git credential helpers
```
pmrowla, line 1: seems like this file was not supposed to be git-committed?

src/scmrepo/git/backend/dulwich/credentials.py
```diff
@@ -86,8 +86,9 @@ def _prepare_command(self) -> Union[str, List[str]]:
             ).strip()
             if shutil.which(executable, path=git_exec_path):
                 executable = os.path.join(git_exec_path, executable)
-
-        return [executable, *argv[1:]]
+        args = [executable]
+        args.extend(map(os.path.expanduser, argv[1:]))
```
pmrowla, line 90: I'm not sure this is safe for us to do. It seems like the credential helper (in this case `git-credential-store`) should be responsible for doing this path expansion and not us, since it should really only be applied to file paths. There could be alternative credential helpers which use `~` in an argument (as a raw character in a username or password for example) and doing this expansion would break that functionality

pmrowla: We may need to double check what cgit actually does here

mgrbyte: Looks like what's happened to me (and what motivated this PR) was that I set a _custom_ file for the credentials.
(caused by reading the git credentials-store docs in haste whilst simultaneously getting flummoxed with trying to get the dvc + git + docker combination working, (but that's another story....)

Seems there's 4 different ways to use git's credentials.helper feature:

1.  Default `~/.git-credentials` used, no URL context.
```bash
git config credentials.helper store
```
scmrepo will find credentials just fine from the default  `~/.git-credentials`  that is not explicitly stored in a  `.gitconfig ` or `.git/config` file.

2.  Custom file, no URL context.
```bash
git config credntials.helper "store --file ${CUSTOM_CREDS_FILE}"
```

3. Default file, with URL context,
```bash
git config credentials.${URL}.username ${USERNAME}
git config credentials.helper store
```

4. Custom file, with URL context.
```
git config credentials.${URL}.username ${USERNAME}
git config credentials.helper "store --file ${CUSTOM_CREDS_FILE}
```

`git config` adds `${CUSTOM_CREDS_FILE}`  _verbatim_  to `~/.gitconfig`  (or `~/.git/config` if  `git config --local` was used).

I believe only in the specific cases where method 2, 4 or if the user writes their .gitconfig by hand _and_ a the path specified starts with `~` does #178 apply where  no credentials are found by the code here.(as the path isn't interpolated in the stack anywhere),

So, in summary what I'd done is more-than-likely a minority issue that few others will run into, and most users using a `store` credential-helper, will probably(?) just use the credentials store command without explicitly using `--file`.

For what it's worth, seems from the git source[code](https://github.com/git/git/blob/c4203212e360b25a1c69467b5a8437d45a373cac/builtin/credential-store.c#L174) that _custom_ paths are appended to a list, and not interpolated.

If this PR warrants further consideration, I agree that the parsing could be more specific, e.g: split `argv[:1]` into tokens and only attempt to invoke `expanduser` on `argv[:1][2]`  in that sequence iif bot `argv[1:][0] == "store"` and  `argv[1:][1] == "--file"` (happy to  look into changing this PR to do that if there's interest).
 
Apologies if this has wasted your bandwidth @pmrowla , at least anyone else making the same mistake and falling down the ensuing rabbit hole will find this and go 🤦‍♂️ 😎 .

pmrowla: Thanks for doing the investigation!

Given that cgit doesn't do the path expansion either, they probably just expect it to be handled by the shell. Would you mind testing a version of this PR that just changes the `subprocess.run` call to use `shell=True` (instead of using `expanduser()` on all of the command args)

mgrbyte: I implemented this locally, passes all tests.

pmrowla: > Attempting to circumvent it, for example by converting the cmd list to a string and using shell=True triggers a CVE warning from bandit.

In this case the bandit warning is expected, and we could suppress it. (Yes, running an arbitrary user-provided shell command is dangerous, but it's also what cgit does in this situation)

pmrowla: Ok, I took another look at the git source, and I think the existing DVC behavior is correct. CGit subcommands are run with `execv` (and not passed through the shell), and for git commands that support path expansion, they are looking for specific args that start with `~` and manually calling their `interpolate_path` function to do the expansion.

In the `git-credential-store` case, they do not do the interpolation (as you already noted), and pass whatever you set in `--file` directly into `fopen()`, so in the #178 case, it will look for a file in your current working dir named `./~/.git-credentials` (so it will look for a directory named `~` in your cwd).

https://github.com/git/git/blob/c4203212e360b25a1c69467b5a8437d45a373cac/builtin/credential-store.c#L173-L180
https://github.com/git/git/blob/c4203212e360b25a1c69467b5a8437d45a373cac/builtin/credential-store.c#L10-L20

pmrowla:
@mgrbyte I'm closing this PR and the original issue for now, see my comment: https://github.com/iterative/scmrepo/pull/180/files#r1121062414

I'll also add a note w/suggested workaround in the original issue

PR #181, opened 2023-02-10
git: Add `rm` method.

daavoo:


src/scmrepo/git/backend/pygit2.py
```diff
@@ -312,6 +312,9 @@ def resolve_commit(self, rev: str) -> "GitCommit":
             [str(parent) for parent in commit.parent_ids],
         )
 
+    def remove(self, paths: Union[str, Iterable[str]], **kwargs):
+        raise NotImplementedError
```
pmrowla, line 316: Would be good to implement this here (porcelain `git rm` equivalent is index remove + local fs remove if the file actually exists in the workspace)

https://www.pygit2.org/index_file.html#pygit2.Index.remove

daavoo: I originally implemented it that way but found some issues. 

When testing as follows:

```
def test_remove(tmp_dir: TmpDir, scm: Git, git: Git):
    tmp_dir.gen({"foo": "foo", "goo": "goo"})
    # scm because pygit2 doesn't implement add/commit
    scm.add(["foo", "goo"])
    scm.commit("add files")
    git.remove(["foo", "goo"])
    staged, *_ = scm.status()
    assert staged["delete"] == ["foo", "goo"]
    assert not (tmp_dir / "foo").exists()
    assert not (tmp_dir / "goo").exists()
```

The `index.remove` implementation failed to pass the test.
I didn't look in depth at why it was the case (should do it if we really want to have `rm` in `pygit2` backend) because I just wanted `rm` in at least 1 backend.

I am not sure if it's just some issue with mixing the `scm` fixtures or if there as deeper problem

pmrowla: One thing I can think of would be that you need to make sure you also call `index.write()` before returning from the pygit2 remove implementation, otherwise other backends won't see the changes (you can see the pygit2 backend implementations for merge and reset for examples)

daavoo: Indeed was missing the `write`

src/scmrepo/git/backend/pygit2.py
```diff
@@ -312,6 +313,18 @@ def resolve_commit(self, rev: str) -> "GitCommit":
             [str(parent) for parent in commit.parent_ids],
         )
 
+    def remove(self, paths: Union[str, Iterable[str]], **kwargs):
+        if isinstance(paths, str):
+            paths = [paths]
+        self.repo.index.remove_all(paths)
+        self.repo.index.write()
+        for path in paths:
+            # TODO: Handle pathspecs
```
daavoo, line 322: We don't seem to handle pathspecs at all in similar cases

tests/test_git.py
```diff
@@ -1070,3 +1071,31 @@ def test_backend_func(
...
+    scm.add(["foo", "goo"])
+    scm.commit("add files")
+    git.remove(["foo", "goo"])
+    staged, *_ = scm.status()
+    assert staged["delete"] == ["foo", "goo"]
+    assert not (tmp_dir / "foo").exists()
+    assert not (tmp_dir / "goo").exists()
+
+
+@pytest.mark.skip_git_backend("dulwich")
```
daavoo, line 1087: Appears that dulwich porcelain remove doesn't work with dirs

src/scmrepo/git/backend/pygit2.py
```diff
@@ -312,6 +313,18 @@ def resolve_commit(self, rev: str) -> "GitCommit":
...
+            if os.path.isfile(path):
+                os.remove(path)
+            elif os.path.isdir(path):
+                rmtree(path, ignore_errors=True)
```
pmrowla, line 326: It's not safe to use `rmtree` here as it will remove any untracked files in that directory tree. This is also why dulwich remove doesn't support dirs and only takes explicit file paths (since the git index does not actually contain directory paths).

Handling directories requires walking the index + workspace trees and diffing them (so we only remove file that exist in both). When you pass a dir into CLI git, it walks that tree prefix to get the explicit file paths, uses those file paths for the index + worktree remove operations, and then checks to see if the parent dir for any removed files is completely empty afterwards (and then does `rmdir` if empty, otherwise it leaves them alone)

Do we need to support a full `rm` from the workspace or can we limit this to only removing from index (equivalent to `git rm --cached`)?

(I guess the question is how are you intending to use this in DVC?)

pmrowla: If you do need workspace remove support, in the pygit2 backend we should have
```py
def remove(..., cached: bool = False):
    if not cached:
        raise NotImplementedError
    ...  # index remove behavior
```

And then handle both the index-only and index+worskpace `cached=True/False` cases in dulwich+gitpython remove.

pmrowla: Also, if we intend to support directory removal we should be requiring `recursive=True` the same as CLI git (but I would suggest we initially only support explicit paths similar to dulwich for this)

daavoo: > (I guess the question is how are you intending to use this in DVC?)

I thought I might have needed it but I don't. 
Ended up adding it for the sake of covering the git API but it is probably not worth it to add it just for the sake of it until it's actually required by some upstream project

PR #196, opened 2023-02-27
fetch_refspecs: Use `dulwich` for ssh urls.

daavoo:
Using pygit2 was causing https://github.com/iterative/dvc/issues/9080

### DIFF
--- a/src/scmrepo/git/__init__.py
+++ b/src/scmrepo/git/__init__.py
@@ -327,12 +327,18 @@ class Git(Base):
         progress: Optional[Callable[["GitProgressEvent"], None]] = None,
         **kwargs,
     ) -> typing.Mapping[str, SyncStatus]:
+        from urllib.parse import urlparse
+
         from .credentials import get_matching_helper_commands
 
-        if "dulwich" in kwargs.get("backends", self.backends.backends) and any(
-            get_matching_helper_commands(url, self.dulwich.repo.get_config_stack())
-        ):
-            kwargs["backends"] = ["dulwich"]
+        if "dulwich" in kwargs.get("backends", self.backends.backends):
+            credentials_helper = any(
+                get_matching_helper_commands(url, self.dulwich.repo.get_config_stack())
+            )
+            parsed = urlparse(url)
+            ssh = parsed.scheme in ("git", "git+ssh", "ssh") or url.startswith("git@")
+            if credentials_helper or ssh:
+                kwargs["backends"] = ["dulwich"]
 
         return self._fetch_refspecs(
             url,


src/scmrepo/git/__init__.py
```diff
@@ -327,12 +327,18 @@ def fetch_refspecs(
...
-        if "dulwich" in kwargs.get("backends", self.backends.backends) and any(
-            get_matching_helper_commands(url, self.dulwich.repo.get_config_stack())
-        ):
-            kwargs["backends"] = ["dulwich"]
+        if "dulwich" in kwargs.get("backends", self.backends.backends):
+            credentials_helper = any(
+                get_matching_helper_commands(url, self.dulwich.repo.get_config_stack())
+            )
+            parsed = urlparse(url)
+            ssh = parsed.scheme in ("git", "git+ssh", "ssh") or url.startswith("git@")
```
pmrowla, line 339: This should maybe be handled here: https://github.com/iterative/scmrepo/blob/e8de01e07edeacae33a9552a1005152b5a153b2a/src/scmrepo/git/backend/pygit2.py#L463-L464

But this may be a different problem entirely if pygit2+SSH is not working on linux/mac. libgit2/pygit2 are supposed to support SSH everywhere except windows.

pmrowla: I guess the issue is that while it supports SSH, libgit2 doesn't parse openssh configs

https://libgit2.org/docs/guides/authentication/

So we probably need to be using `pygit2.RemoteCallbacks` and loading the keyfile or user/pass ourselves (we can use asyncssh for this)

pmrowla (approved):
This is ok as a workaround until we handle libgit2/pygit2 auth callbacks properly

src/scmrepo/git/__init__.py
```diff
@@ -327,12 +327,18 @@ def fetch_refspecs(
...
+
         from .credentials import get_matching_helper_commands
 
-        if "dulwich" in kwargs.get("backends", self.backends.backends) and any(
-            get_matching_helper_commands(url, self.dulwich.repo.get_config_stack())
-        ):
-            kwargs["backends"] = ["dulwich"]
+        if "dulwich" in kwargs.get("backends", self.backends.backends):
+            credentials_helper = any(
+                get_matching_helper_commands(url, self.dulwich.repo.get_config_stack())
```
skshetry, line 336: IIRC credential-helper is not supposed to work for ssh, only for http urls. So this should be fixed in `get_matching_helper_commands()` or in upstream dulwich.

Regarding ssh not working in pygit2, let's investigate/fix it separately.

daavoo: > IIRC credential-helper is not supposed to work for ssh, only for http urls. So this should be fixed in get_matching_helper_commands() or in upstream dulwich.

Not sure I understand what you mean. The previous logic is preserved but the change is to also force `dulwich` if the URL is ssh

daavoo:
> This is ok as a workaround until we handle libgit2/pygit2 auth callbacks properly

Opened https://github.com/iterative/scmrepo/issues/197

ReveStobinson:
Just manually installed the `0.1.13` release with `dvc=2.45.1`, and these changes appear to resolve https://github.com/iterative/dvc/issues/9080 for me! Thank you, all, for the quick response on this!

src/scmrepo/git/__init__.py
skshetry: `get_matching_helper_commands` should not match anything for ssh urls. Please try to address comments before merging.

daavoo: > get_matching_helper_commands should not match anything for ssh urls, so it should not enter this block.

Sorry, I am still not sure I follow. You mean it should be something like:

```python
if ssh or any(get_matching_helper_commands(url, self.dulwich.repo.get_config_stack()):
```

PR #200, opened 2023-03-03
git: Add `get_remote_url` method.

daavoo:


pmrowla:
I commented in the DVC PR, but this shouldn't be necessary (and scmrepo git remote commands should work with both raw URLs and git remote names). Backend specific remote functions should raise `NotImplementedError` when we try to use them with an unsupported remote type, so `pygit2._fetch_refspecs` should be raising `NotImplementedError` when we try to use it with SSH

https://github.com/iterative/dvc/pull/9106#discussion_r1124323285

PR #210, opened 2023-03-17
git: prefer gitpython for auth ops

### LINKED ISSUE #208: clone: use gitpython's CLI git first and fallback to dulwich if CLI git is not available
With all of the auth problems and stuff like https://github.com/iterative/dvc-ssh/issues/20 , we are probably better off just using cli git for clone first and fallback to dulwich if not available.

clone used to be the last gitpython thing for a very long time without any problems and the whole point of migrating clone to dulwich was so that dvc get/import/list could work in environments without git cli (e.g. during deployment in docker images), but we could get both benefits by just falling back to dulwich.

CC @dberenbaum

efiop:
clone/push/pull/fetch/push_refspecs/fetch_refspecs all rely on auth heavily and git CLI, unfortunately, handles it the best across the board. So we will prefer using it for those operations but if CLI git is not available - we'll fall back to dulwich/pygit2.

Note that push/fetch_refspecs are not implemented in gitpython backend yet, so this won't work for them until those are implemented.

Fixes #208

### DIFF
--- a/src/scmrepo/git/__init__.py
+++ b/src/scmrepo/git/__init__.py
@@ -3,12 +3,11 @@
 import logging
 import os
 import re
-import typing
 from collections import OrderedDict
 from collections.abc import Mapping
 from contextlib import contextmanager
 from functools import partialmethod
-from typing import TYPE_CHECKING, Callable, Dict, Iterable, Optional, Tuple, Type, Union
+from typing import Dict, Iterable, Optional, Tuple, Type, Union
 
 from funcy import cached_property, first
 from git import GitCommandNotFound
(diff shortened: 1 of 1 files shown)


efiop:
@pmrowla @skshetry Folks, really need your opinion and review here. 🙏

pmrowla (approved):
I would still prefer that we just spend the time fixing these auth problems in pygit/dulwich instead of re-introducing default use of gitpython

efiop:
@pmrowla I agree with you. Do you think it is doable short term? I could jump into it. Things like https://github.com/iterative/dvc-ssh/issues/20 look concerning though.

pmrowla:
Porting the apple openssh to asyncssh is non-trivial and probably not what we want. What we should be doing in dulwich is checking for unsupported configuration options in our asyncssh vendor (like `UseKeychain` which specifies using the apple keychain behavior) and then fallback to the default dulwich SSH vendor instead (which just calls system `ssh` via subprocess, and would use apple's openssh build and work as expected).

We default to using asyncssh since it removes a dependency on `ssh`, which is not always available in windows (or in containers)

pmrowla:
The other issues are mostly about implementing auth callbacks for pygit2, which should be straightforward

skshetry (changes requested):
We should implement `push_refspecs`/`fetch_refspecs` in GitPython too if we make this change. Using a different backend for cloning and a different backend for push_refspecs will be painful (which it already is). I'd prefer to have this change together.

efiop:
@pmrowla Sounds reasonable. Let's discuss during planning today.

PR #216, opened 2023-03-24
git.credentials: implement full fill/approve/reject workflow

### LINKED ISSUE #215: credential-store: support `store` for new credentials
# Bug Report

## Description

When using the [git credential manager](https://github.com/GitCredentialManager/git-credential-manager) to manage login information and using https-based remotes in dvc (e.g., `dvc import https://github.com/...`), the credentials are not properly cached. If I just try to fetch some repo using https, GCM correctly uses the saved credentials. In the case of dvc, the login window appears for every step of the pipeline, meaning I have to sign in three times for a single import. GitHub already warned me that GCM triggered some rate limit for my account because of this. Here is an exemplary terminal output:

```console
Importing '/ (https://github.com/xxx/xxx.git)' -> 'xxx'
Cloning xxx.git|                                                                                                                                                                                                                  |0.00/? [00:00,      ?obj/s]
info: please complete authentication in your browser...
Cloning xxx.git|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| Compressing |876/876 [03:38,    219s/obj]
info: please complete authentication in your browser...
info: please complete authentication in your browser...
```

The same happens for subsequent runs of `dvc pull`.


### Reproduce

1. dvc import https://github.com/xxx/xxx.git
2. sign in with GitHub in the GCM window
3. sign in again
4. ...



### Expected

Credentials are used without the dialogue popping up at all (just like it does for the regular git commands).


### Environment information


**Output of `dvc doctor`:**

```console
DVC version: 2.43.1 (brew)
---------------------------------
Platform: Python 3.11.1 on macOS-13.2-x86_64-i386-64bit
Subprojects:
	dvc_data = 0.35.1
	dvc_objects = 0.19.0
	dvc_render = 0.0.17
	dvc_task = 0.1.11
	dvclive = 1.3.3
	scmrepo = 0.1.6
Supports:
	azure (adlfs = 2023.1.0, knack = 0.10.1, azure-identity = 1.12.0),
	gdrive (pydrive2 = 1.15.0),
	gs (gcsfs = 2023.1.0),
	http (aiohttp = 3.8.3, aiohttp-retry = 2.8.3),
	https (aiohttp = 3.8.3, aiohttp-retry = 2.8.3),
	oss (ossfs = 2021.8.0),
	s3 (s3fs = 2023.1.0, boto3 = 1.24.59),
	ssh (sshfs = 2023.1.0),
	webdav (webdav4 = 0.9.8),
	webdavs (webdav4 = 0.9.8),
	webhdfs (fsspec = 2023.1.0)
Cache types: reflink, hardlink, symlink
Cache directory: apfs on /dev/disk1s1s1
Caches: local
Remotes: None
Workspace directory: apfs on /dev/disk1s1s1
Repo: dvc, git
```

pmrowla:
- Adds higher level `git.credentials` API equivalent to `git-credentials`
    - `scmrepo` clients should use `Credential.fill/approve/reject` rather than using `Helper.get/store/erase` directly (as described in https://git-scm.com/docs/git-credential#_typical_use_of_git_credential)
    - Microsoft's GCM now fully supported
- Currently only used in dulwich backend for HTTP clients
- Credentials that are successfully filled by helper backend will be approved/stored upon a successful request

Closes #215

### DIFF
--- a/src/scmrepo/git/backend/dulwich/client.py
+++ b/src/scmrepo/git/backend/dulwich/client.py
@@ -1,10 +1,8 @@
-# Temporarily added while waiting for upstream PR to be merged.
-# See https://github.com/jelmer/dulwich/pull/976
+from typing import Optional
 
 from dulwich.client import Urllib3HttpGitClient
-from dulwich.config import StackedConfig
 
-from scmrepo.git.credentials import CredentialNotFoundError, get_credentials_from_helper
+from scmrepo.git.credentials import Credential, CredentialNotFoundError
 
 
 class GitCredentialsHTTPClient(Urllib3HttpGitClient):  # pylint: disable=abstract-method
@@ -24,17 +22,23 @@ class GitCredentialsHTTPClient(Urllib3HttpGitClient):  # pylint: disable=abstrac
             **kwargs,
         )
 
+        self._store_credentials: Optional["Credential"] = None
         if not username:
-            try:
-                helper_username, helper_password = get_credentials_from_helper(
-                    base_url, config or StackedConfig.default()
-                )
-            except CredentialNotFoundError:
-                pass
-            else:
-                credentials = helper_username + b":" + helper_password
-                import base64
+            import base64
 
-                encoded = base64.b64encode(credentials).decode("ascii")
-                basic_auth = {"authorization": f"Basic {encoded}"}
-                self.pool_manager.headers.update(basic_auth)
+            try:
+                creds = Credential(url=base_url).fill()
+            except CredentialNotFoundError:
+                return
+            encoded = base64.b64encode(
+                f"{creds.username}:{creds.password}".encode()
+            ).decode("ascii")
+            basic_auth = {"authorization": f"Basic {encoded}"}
+            self.pool_manager.headers.update(basic_auth)
+            self._store_credentials = creds
+
+    def _http_request(self, *args, **kwargs):
+        result = super()._http_request(*args, **kwargs)
+        if self._store_credentials is not None:
+            self._store_credentials.approve()
+        return result
--- a/src/scmrepo/git/credentials.py
+++ b/src/scmrepo/git/credentials.py
@@ -26,21 +26,31 @@ https://git-scm.com/book/en/v2/Git-Tools-Credential-Storage
 Currently Dulwich supports only the `get` operation
 
 """
+import locale
+import logging
 import os
 import shlex
 import shutil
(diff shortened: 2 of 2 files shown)


dberenbaum:
Thanks @pmrowla! Does it have any impact on prompting/interactive auth? Previously, I don't think we supported it at all, but AFAIK `git credential fill` may prompt users if there are no credentials found. Does it mean that dulwich HTTP git clients will prompt for missing credentials?

pmrowla:
> Thanks @pmrowla! Does it have any impact on prompting/interactive auth? Previously, I don't think we supported it at all, but AFAIK `git credential fill` may prompt users if there are no credentials found. Does it mean that dulwich HTTP git clients will prompt for missing credentials?

If the credential manager requires interactive input then yes. With microsoft's standard GCM it opens a browser window for http logins. This was already the scmrepo behavior prior to this patch though (since the original credential helper support was added).

The difference was that we did not actually save the credentials afterwards, so users would be prompted every time we did a git operation that required auth, whereas now they are prompted on the first call and then we save the credentials in GCM, the same way that CLI git handles it.

dberenbaum:
Sorry @pmrowla, I think I was conflating the credential helpers with the default git behavior for [requesting credentials](https://git-scm.com/docs/gitcredentials#_requesting_credentials). That is really only useful initially, but this is one place where CLI Git doesn't fail and DVC does. What do you think about supporting something like this?

pmrowla:
We could support interactive prompting but it's probably not a quick fix since dulwich doesn't have native auth callback support yet. The intended workflow is that git should try an unauthenticated request (without using any credential helper or prompting) and then if it fails it should retry with helpers and/or prompting.

The current behavior in scmrepo is that we try credential helpers first if they exist. Technically this means we load GCM credentials even for a request that would not require it, like cloning a public repo, but in practice this isn't really an issue for the credential helper case. For the interactive CLI prompt it does make a difference, since with the current behavior we would have to prompt the user before making every request (since whatever the user enters via the interactive prompt is not saved).

We can implement this in scmrepo and then patch it into dulwich later, but it probably won't be a quick/small change.

dberenbaum:
What do you think about falling back to Git CLI when an auth error is raised?

pmrowla:
@dberenbaum I would prefer that we just put the time into implementing it instead of using the CLI fallback

PR #226, opened 2023-03-31
credentials: generalize prompt behavior

### LINKED ISSUE #225: git-credential follow ups
- [x] support `GIT_ASKPASS`/`core.askPass`/`SSH_ASKPASS`
- [x] support `quit=true/1` to short-circuit additional credential checks
- [x] drop `memory_only` flag (after testing, CLI git does send prompted credentials to existing helpers like GCM)

pmrowla:
closes #225

### DIFF
--- a/src/scmrepo/git/credentials.py
+++ b/src/scmrepo/git/credentials.py
@@ -68,6 +68,10 @@ class CredentialNotFoundError(SCMError):
     """Error occurred while retrieving credentials/no credentials available."""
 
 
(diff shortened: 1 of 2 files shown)


pmrowla:
askpass support looks like this:

https://user-images.githubusercontent.com/651988/229072577-3a1af2a3-5b2e-4c2c-bddc-a43251247b0e.mp4

(example uses dvc commands in the vscode terminal, but it would be the same if you ran an experiment that had an import in the pipeline)

PR #235, opened 2023-04-18
git-credential: disable prompt when not attached to a tty

### LINKED ISSUE #234: credentials: disable interactive prompt if not in TTY
see https://github.com/iterative/dvc/issues/9339

pmrowla:
Fixes #234

### DIFF
--- a/src/scmrepo/git/credentials.py
+++ b/src/scmrepo/git/credentials.py
@@ -283,6 +283,8 @@ def _input_tty(prompt: str = "Username: ") -> str:
     from contextlib import ExitStack
 
     if os.name == "nt":
+        if not sys.stdin.isatty():
+            raise EOFError
         return input(prompt)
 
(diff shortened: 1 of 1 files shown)


src/scmrepo/git/credentials.py
```diff
@@ -283,6 +283,8 @@ def _input_tty(prompt: str = "Username: ") -> str:
     from contextlib import ExitStack
 
     if os.name == "nt":
+        if not sys.stdout.isatty():
```
skshetry, line 286: We should be looking for `sys.stdin.isatty()` not `sys.stdout`, no?

pmrowla: Hmm, I guess technically that's probably more correct, but I'm not sure we should be trying to read from stdin if we can't actually write out the prompt either?

PR #244, opened 2023-06-14
handle empty credential helper

dberenbaum:
Currently fails with a config like:

```
[credential]
	helper = 
```

I was using that to debug other auth issues by resetting my credential helpers and encountered an error.


efiop:
Great catch!

pmrowla (approved):

PR #255, opened 2023-07-20
dulwich: support interactive SSH key passphrase prompt in asyncssh vendor

pmrowla:
Closes https://github.com/iterative/dvc/issues/9737

- Makes scmrepo behave the same as git when prompting for a passphrase

```
$ dvc import git@github.com:pmrowla/test-dvc-private.git data/data.xml
Importing 'data/data.xml (git@github.com:pmrowla/test-dvc-private.git)' -> 'data.xml'
Cloning test-dvc-private.git|                             |0.00/? [00:00,      ?obj/s]
Enter passphrase for key '~/git/scratch/test-ssh-passphrase/id_ed25519':

To track the changes with git, run:

        git add data.xml.dvc

To enable auto staging, run:

        dvc config core.autostage true
```

### DIFF
--- a/src/scmrepo/git/backend/dulwich/asyncssh_vendor.py
+++ b/src/scmrepo/git/backend/dulwich/asyncssh_vendor.py
@@ -5,10 +5,12 @@ from typing import (
     TYPE_CHECKING,
     Callable,
     Coroutine,
+    Dict,
     Iterator,
     List,
     Optional,
     Sequence,
+    cast,
 )
 
 from asyncssh import SSHClient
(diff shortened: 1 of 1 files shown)


pmrowla:
We should also probably consider using a similar `InteractiveSSHClient` implementation in dvc-ssh instead of using the current `ask_password`/`ask_passphrase` options (which require the user to use those remote flags and prompt before we try to make any SSH connection or try any specific SSH keys)

related: https://github.com/iterative/dvc-ssh/pull/3

pmrowla:
On the scmrepo side, some of this could also be generalized a bit more and re-used in `git.credentials` to implement full support for SSH auth in pygit2/libgit2 (which will require loading keys via asyncssh and then passing them into the pygit2 credential callbacks), but that can be done in a follow-up PR if/when needed

efiop (approved):

PR #268, opened 2023-08-20
feat(pygit2): author and committer don't break if env is set and git conf is not

shcheklein:
Fixes https://github.com/iterative/studio/issues/7185

PyGit2 (and our backend in scmrepo) throws an exception if there is no Git config set. Might be related to this https://github.com/libgit2/libgit2/issues/1528

Here we are trying to get env vars first, if they are specified and don't fail fast on the default signature call.

### DIFF
--- a/src/scmrepo/git/backend/pygit2/__init__.py
+++ b/src/scmrepo/git/backend/pygit2/__init__.py
@@ -142,14 +142,32 @@ class Pygit2Backend(BaseGitBackend):  # pylint:disable=abstract-method
     def _get_signature(self, name: str) -> "Signature":
         from pygit2 import Signature
 
-        sig = self.default_signature
+        try:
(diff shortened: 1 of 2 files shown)


pmrowla (approved):
This LGTM

I'm assuming, it's still marked as draft pending studio testing?

shcheklein:
Actually I wanted to add test here. Didn't have enough time today :(

shcheklein:
And yes, wanted to test with Studio locally + I think we'll need to have a Studio test (can be followup)

shcheklein:
@pmrowla  Should be ready to review. I've done basic local testing with Studio. We can double check on dev env when the new scmrepo version lands. Also on the Studio side we need an E2E test for MR actions (can be a followup).

pmrowla (approved):

PR #269, opened 2023-08-22
gitpython: raise NotImplementedError when import fails

pmrowla:
see: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=766627&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=986b1512-c876-5f92-0d81-ba851554a0a3

### DIFF
--- a/src/scmrepo/git/__init__.py
+++ b/src/scmrepo/git/__init__.py
@@ -11,7 +11,6 @@ from functools import partialmethod
 from typing import TYPE_CHECKING, Callable, Dict, Iterable, Optional, Tuple, Type, Union
 
 from funcy import cached_property, first
-from git import GitCommandNotFound
 from pathspec.patterns import GitWildMatchPattern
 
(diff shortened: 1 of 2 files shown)


src/scmrepo/git/__init__.py
```diff
@@ -11,7 +11,6 @@
 from typing import TYPE_CHECKING, Callable, Dict, Iterable, Optional, Tuple, Type, Union
 
 from funcy import cached_property, first
-from git import GitCommandNotFound
```
pmrowla, line 14: gitpython raises an `ImportError` the first time it tries to find CLI git and it is unavailable, `GitCommandNotFound` is only raised on subsequent attempts to find CLI git. gitpython import should be lazy, otherwise `import scmrepo.git` will fail if CLI git is unavailable

Both the `ImportError` and `GitCommandNotFound` should be handled within the gitpython backend and not at this level

PR #278, opened 2023-10-03
support LFS smudge (read operations)

pmrowla:
Will close https://github.com/iterative/dvc/issues/9175
Requires https://github.com/libgit2/pygit2/pull/1237

- Adds minimal read-only LFS implementation
- `lfs smudge` filtering is enabled by via libgit2/pygit2 backend filter
    - When Git checks out (or reads) a blob with the `filter=lfs` .gitattribute the LFS object will be read from local storage or fetched (per default CLI `lfs smudge` behavior)
- `lfs fetch` is implemented via `scmrepo.git.lfs.fetch` and can be used to fetch multiple LFS objects in a batch operation (so subsequent git checkout/blob read can be done from local storage)
- Minimal CLI implementation is available for some commands. This is mainly useful for compatibility testing purposes and should not be considered a complete replacement for CLI LFS, commands only implement the typical default usage flags.:
    - `python -m scmrepo.git.lfs.fetch` (`git lfs fetch`)
    - `python -m scmrepo.git.lfs.pointer` (`git lfs pointer`)
    - `python -m scmrepo.git.lfs.smudge` (`git lfs smudge`)
- [LFS configuration](https://github.com/git-lfs/git-lfs/blob/main/docs/man/git-lfs-config.adoc) is only partially implemented. Currently supported options (`.lfsconfig` is supported in addition to Git repo config when applicable):
    - `lfs.storage`
    - `lfs.url`
    - `remote.lfsdefault`
    - `remote.*.lfsurl`
- `lfs.fetchinclude` and `lfs.fetchexclude` equivalent behavior is supported internally via `scmrepo.git.lfs.fetch(include=..., exclude=...)`, but `fetch()` will not read the configured include/exclude patterns by default).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,6 +70,7 @@ files = ["src", "tests"]
 [[tool.mypy.overrides]]
 module = [
     "pygtrie",
+    "dvc_http.*",
     "funcy",
     "git",
     "gitdb.*",
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ packages = find:
 install_requires=
     gitpython>3
     dulwich>=0.21.6
-    pygit2>=1.13.0
+    pygit2>=1.13.3
     pygtrie>=2.3.2
     fsspec>=2021.7.0
     pathspec>=0.9.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,6 +33,8 @@ install_requires=
     asyncssh>=2.13.1,<3
     funcy>=1.14
     shortuuid>=0.5.0
+    dvc-objects>=1.0.1,<2
+    dvc-http>=2.29.0
 
 [options.extras_require]
 tests =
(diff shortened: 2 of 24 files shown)


pmrowla:
Current state of the PR is that `GitFilesystem.open()` will return filtered content for lfs-tracked files when they are available locally in `.git/lfs/objects/...` (remote fetch is not implemented yet)

in a repo with lfs-tracked file:
```
$ git lfs ls-files
b5bb9d8014 * foo.lfs
```

test script:
```python 
from scmrepo.git import Git

scm = Git()
try:
    fs = scm.get_fs("HEAD")
    with fs.open("foo.lfs") as f:
        print(f.read().decode("utf-8"))
finally:
    scm.close()
```

current `main` (we read the raw git object containing the LFS pointer file):
```
$ python test.py
version https://git-lfs.github.com/spec/v1
oid sha256:b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c
size 4
```

with this PR (we load the LFS pointer file and the read the actual `.git/lfs/objects/b5/bb/...` file):
```
$ python test.py
foo

```

efiop:
New pygit2 is out https://pypi.org/project/pygit2/ 🎉

PR #283, opened 2023-11-10
Convert check_ref_format to class method

amritghimire:
Refactored the check_ref_format function in dulwich/__init__.py to a class method for more flexibility and to allow calling it without instantiating the class. This simplifies the usage of this method while maintaining existing functionality.

More context for this is at https://github.com/iterative/studio/pull/8187#issuecomment-1804589654

### DIFF
--- a/src/scmrepo/git/backend/base.py
+++ b/src/scmrepo/git/backend/base.py
@@ -406,8 +406,9 @@ class BaseGitBackend(ABC):
     def get_remote_url(self, remote: str) -> str:
         """Return URL for the specified remote."""
 
+    @staticmethod
     @abstractmethod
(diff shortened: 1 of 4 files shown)


pmrowla:
If we want to make this a classmethod, it should be a classmethod in the base backend abstract class and in all of the implementations, and top level `scmrepo.git.Git` would really also need a way to call an appropriate backend classmethod (so the scmrepo user/caller should only need to do `Git.check_ref_format` and then it would pick an appropriate `Backend.check_ref_format` to use)

But honestly IMO it's not a problem for studio to just call `dulwich.refs.check_ref_format` directly in this situation

amritghimire:
Closing this for now. @shcheklein Feel free to takeover if you think this is necessary.

PR #290, opened 2023-12-13
lfs: replace wrap_and_branch_coro with equivalent

skshetry:


skshetry:
It's unclear to me why we are even using `dvc-http` and `dvc-objects` here.

skshetry:
I'm running with this to unblock https://github.com/iterative/dql/pull/1058.

pmrowla:
> It's unclear to me why we are even using `dvc-http` and `dvc-objects` here.

It's so we can reuse the logic for batched transfers and the aiohttp large download timeouts. I didn't see the point in copy/pasting everything from dvc-http and dvc-objects generic transfer into scmrepo

skshetry:
dvc-http is a dvc's plugin. I have fixed the immediate issue, but we are creating a circular dependency issue here with this.

Maybe we can add support for batched transfers for fsspec filesystems in dvc-objects?
The aiohttp large download timeouts fixes are small that we can copy it here as well (or upstream it to fsspec).

Ideally, the fix would be to not have our own wrappers.

PR #315, opened 2024-01-29
lfs: use `HTTPFileSystem` from `fsspec` instead of `dvc-http`

sisp:
I've refactored the `LFSClient` class some more to directly use `fsspec`'s `HTTPFileSystem` implementation instead of the one from `dvc-http`, as most of `dvc-http`'s implementation is irrelevant to making HTTP requests independent of DVC. I've retained the retry client, which `dvc-http` also uses, because that's generally consistent with the official Git LFS client's implementation. In fact, the [official Git LFS client also supports retries for Batch API requests](https://github.com/git-lfs/git-lfs/blob/82f58ad228ff49abb31618152b1992820bb84f5e/tq/api.go#L94), which are disabled by `dvc-http`'s `ReadOnlyRetryClient` class. To otherwise keep the retry behavior as before, I've [copied the retry settings from `dvc-http`](https://github.com/iterative/dvc-http/blob/1df9de067248c9408b5e8b24c53c3f6998dd141c/dvc_http/__init__.py#L42-L44) including the [hardcoded number of parallel jobs inherited from `dvc_objects.fs.base.FileSystem`](https://github.com/iterative/dvc-objects/blob/76eed478b0351dab5c0c150e23cad6ff41e47b57/src/dvc_objects/fs/base.py#L75). In a follow-up PR, I believe we should make this value configurable and get the `jobs` value from DVC's used `FileSystem` instance in [`dvc.scm.lfs_prefetch`](https://github.com/iterative/dvc/blob/196c1c7a17d15e50be5da359599c7e51059cec30/dvc/scm.py#L269-L293) like this:

```diff
 def lfs_prefetch(fs: "FileSystem", paths: List[str]):
     from scmrepo.git.lfs import fetch as _lfs_fetch

     from dvc.fs.dvc import DVCFileSystem
     from dvc.fs.git import GitFileSystem

     if isinstance(fs, DVCFileSystem) and isinstance(fs.repo.fs, GitFileSystem):
         git_fs = fs.repo.fs
         scm = fs.repo.scm
         assert isinstance(scm, Git)
     else:
         return

     try:
         if "filter=lfs" not in git_fs.open(".gitattributes").read():
             return
     except OSError:
         return
     with TqdmGit(desc="Checking for Git-LFS objects") as pbar:
         _lfs_fetch(
             scm,
             [git_fs.rev],
             include=[(path if path.startswith("/") else f"/{path}") for path in paths],
             progress=pbar.update_git,
+            jobs=fs.jobs,
         )
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
     "funcy>=1.14",
     "shortuuid>=0.5.0",
     "dvc-objects>=4,<5",
-    "dvc-http>=2.29.0",
+    "aiohttp-retry>=2.5.0",
 ]
 
 [project.urls]
@@ -109,7 +109,6 @@ files = ["src", "tests"]
 [[tool.mypy.overrides]]
 module = [
     "pygtrie",
-    "dvc_http.*",
     "funcy",
     "git",
     "gitdb.*",
--- a/src/scmrepo/git/lfs/client.py
+++ b/src/scmrepo/git/lfs/client.py
@@ -6,10 +6,9 @@ from typing import TYPE_CHECKING, Any, Optional
 
 import aiohttp
 from aiohttp_retry import ExponentialRetry, RetryClient
-from dvc_objects.executors import batch_coros
 from dvc_objects.fs import localfs
 from dvc_objects.fs.utils import as_atomic
-from fsspec.asyn import sync_wrapper
+from fsspec.asyn import _run_coros_in_chunks, sync_wrapper
 from fsspec.callbacks import DEFAULT_CALLBACK
 from fsspec.implementations.http import HTTPFileSystem
 from funcy import cached_property
@@ -178,7 +177,7 @@ class LFSClient(AbstractContextManager):
             headers = download.get("header", {})
             to_path = storage.oid_to_path(obj.oid)
             coros.append(_get_one(url, to_path, headers=headers))
-        for result in await batch_coros(
+        for result in await _run_coros_in_chunks(
             coros, batch_size=self._JOBS, return_exceptions=True
         ):
             if isinstance(result, BaseException):
--- a/src/scmrepo/git/lfs/client.py
+++ b/src/scmrepo/git/lfs/client.py
@@ -1,15 +1,17 @@
 import logging
 from collections.abc import Iterable
 from contextlib import AbstractContextManager
+from multiprocessing import cpu_count
 from typing import TYPE_CHECKING, Any, Optional
 
 import aiohttp
-from dvc_http import HTTPFileSystem
+from aiohttp_retry import ExponentialRetry, RetryClient
 from dvc_objects.executors import batch_coros
 from dvc_objects.fs import localfs
 from dvc_objects.fs.utils import as_atomic
 from fsspec.asyn import sync_wrapper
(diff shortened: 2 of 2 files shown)


sisp:
The cyclic import seems to be an upstream problem in `dvc-objects`:

```console
$ python -c 'from dvc_objects.executors import batch_coros'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".venv/lib/python3.11/site-packages/dvc_objects/executors.py", line 21, in <module>
    from .fs.callbacks import Callback
  File ".venv/lib/python3.11/site-packages/dvc_objects/fs/__init__.py", line 5, in <module>
    from . import generic  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/dvc_objects/fs/generic.py", line 11, in <module>
    from dvc_objects.executors import ThreadPoolExecutor, batch_coros
ImportError: cannot import name 'ThreadPoolExecutor' from partially initialized module 'dvc_objects.executors' (most likely due to a circular import) (.venv/lib/python3.11/site-packages/dvc_objects/executors.py)
```

I'm looking into it.

sisp:
We could use a slightly ugly workaround to enforce a different import order in `scmrepo/git/lfs/client.py`:

```diff
 import aiohttp
 from aiohttp_retry import ExponentialRetry, RetryClient
-from dvc_objects.executors import batch_coros
 from dvc_objects.fs import localfs
 from dvc_objects.fs.callbacks import DEFAULT_CALLBACK
 from dvc_objects.fs.utils import as_atomic
@@ -14,6 +13,11 @@ from fsspec.asyn import sync_wrapper
 from fsspec.implementations.http import HTTPFileSystem
 from funcy import cached_property
 
+if True:
+    # NOTE: This is a workaround to avoid an import cycle. Importing from
+    # `dvc_objects.executors` must come after importing from `dvc_objects.fs`.
+    from dvc_objects.executors import batch_coros
+
 from scmrepo.git.credentials import Credential, CredentialNotFoundError
 
 from .exceptions import LFSError
```

Or we could do a lazy import `batch_coros` in `LFSClient._download()`.

WDYT, @pmrowla?

pmrowla:
> Or we could do a lazy import of `batch_coros` in `LFSClient._download()`.

It would be better to drop the dependency on dvc-objects entirely and use the fsspec version of `batch_coros`. The dvc-objects implementation is more optimized, but it's not a big issue in this use case, and it's been noted elsewhere that scmrepo shouldn't really import any of the DVC libraries since it introduces the circular dependencies.

pmrowla:
> In a follow-up PR, I believe we should make this value configurable and get the `jobs` value from DVC's used `FileSystem` instance in [`dvc.scm.lfs_prefetch`](https://github.com/iterative/dvc/blob/196c1c7a17d15e50be5da359599c7e51059cec30/dvc/scm.py#L269-L293) like this:

scmrepo can probably just get rid of the DVC `jobs`/`_JOBS` settings entirely and just pass `batch_size` into the fsspec operations (and fallback to fsspec's default batch size which is also based on cpu count), but as you noted this can be done in a follow up PR

pmrowla (approved):

PR #320, opened 2024-02-06
lfs: drop dvc objects dependency

pmrowla:
- Drops dvc-objects usage in favor of fsspec equivalents
- Drops DVC jobs/_JOBS usage in favor of exposing fsspec batch_size (see https://github.com/iterative/scmrepo/pull/315#issuecomment-1928855658)

### DIFF
--- a/src/scmrepo/git/lfs/progress.py
+++ b/src/scmrepo/git/lfs/progress.py
@@ -1,9 +1,112 @@
-from typing import BinaryIO, Callable, Optional, Union
+import logging
+import sys
+from typing import Any, BinaryIO, Callable, ClassVar, Optional, Union
 
-from dvc_objects.fs.callbacks import TqdmCallback
-from fsspec.callbacks import DEFAULT_CALLBACK, Callback
+from fsspec.callbacks import DEFAULT_CALLBACK, Callback, TqdmCallback
+from tqdm import tqdm
 
 from scmrepo.progress import GitProgressEvent
+
+
+class _Tqdm(tqdm):
+    """
+    maximum-compatibility tqdm-based progressbars
+    """
+
+    BAR_FMT_DEFAULT = (
+        "{percentage:3.0f}% {desc}|{bar}|"
+        "{postfix[info]}{n_fmt}/{total_fmt}"
+        " [{elapsed}<{remaining}, {rate_fmt:>11}]"
+    )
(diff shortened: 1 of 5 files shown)


src/scmrepo/git/lfs/progress.py
```diff
@@ -1,11 +1,114 @@
...
 
-from dvc_objects.fs.callbacks import TqdmCallback
-from fsspec.callbacks import DEFAULT_CALLBACK, Callback
+from fsspec.callbacks import DEFAULT_CALLBACK, Callback, TqdmCallback
+from tqdm import tqdm
 
 from scmrepo.progress import GitProgressEvent
 
 
+class _Tqdm(tqdm):
```
skshetry, line 11: @pmrowla, do you think it's possible to use `dvc`'s `TqdmCallback` and `Tqdm` implementation?

pmrowla: It should be possible, but I do think there's value in scmrepo being able to handle reporting progress entirely separately from DVC

This `_Tqdm` subclass isn't strictly necessary in scmrepo, but the default fsspec TqdmCallback and default tqdm progressbars don't handle the branched nesting particularly well.

pmrowla: Yeah that should work fine

skshetry: @pmrowla, is it just the `lfs_fetch`/`fetch` that should take progressbar from dvc? I see that the function is being called in many different places like `open()` and `smudge`/`lfsfilter` indirectly.

I guess there's no way to pass the callbacks to them, right?

pmrowla: From DVC's perspective it only matters for the high level fetch. Fetch can still be done from open/smudge because that's how the LFS CLI works.

You can manually do an explicit `git lfs fetch` to fetch LFS objects (which is the equivalent of what DVC does), and progress for the fetch will get reported similar to how `git fetch` progress is normally reported. But if you are doing a `git checkout` and there is an LFS object that has not been fetched that needs to be checked out, git + LFS will do the underlying LFS fetch for that object at that time, but it will just look like a slow/hanging checkout.
(You can configure LFS to write progress to a specific file instead of stderr/stdout for these situations where there is no equivalent CLI `git` progress reporting, but that's not implemented in scmrepo)

PR #325, opened 2024-02-27
lfs: add support for Git SSH URLs

### LINKED ISSUE #288: lfs: support SSH `git-lfs-authenticate`
This is more applicable for custom/private LFS instances than the usual git forges, but LFS uses a custom server-side helper to obtain credentials for SSH git servers that do not use HTTP git credentials

See: https://github.com/git-lfs/git-lfs/blob/main/docs/api/authentication.md#ssh

sisp:
I've added support for Git SSH URLs to the Git LFS client. Resolves #288.

I've refactored the `LFSClient` class a little to have the common code in `LFSClient` and implement logic specific to Git HTTP URLs and Git SSH URLs in subclasses. The `LFSClient` class is now abstract with one abstract method `_get_auth_header(upload: bool): dict` which gets implemented by the private subclasses `_HTTPLFSClient` and `_SSHLFSClient`. `_HTTPLFSClient` retrieves credentials as before using the credential helper, but instead of using AIOHTTP's `auth` argument, it creates the basic auth header manually using `BasicAuth(...).encode()` which gets passed along with other headers. `_SSHLFSClient` retrieves credentials using `ssh git@<host> <project_path> upload|download` (as documented in the [Git LFS SSH authentication docs](https://github.com/git-lfs/git-lfs/blob/main/docs/api/authentication.md#ssh)) and takes the `header` dict from its response, which contains authentication data.

Some questions regarding this implementation:
* Is using `_get_ssh_vendor` from `scmrepo.git.backend.dulwich` the right way to create an SSH client?
* Any suggestions about testing? I don't feel too comfortable adding this feature without tests, but I'm not sure what the best approach to testing would be. Perhaps using a Git LFS test server? I've manually tested both Git SSH URLs and Git HTTP URLs of public and private projects, all scenarios are working for me.

### DIFF
--- a/src/scmrepo/git/lfs/client.py
+++ b/src/scmrepo/git/lfs/client.py
@@ -1,6 +1,9 @@
+import json
 import logging
 import os
+import re
 import shutil
+from abc import abstractmethod
 from collections.abc import Iterable, Iterator
 from contextlib import AbstractContextManager, contextmanager, suppress
 from tempfile import NamedTemporaryFile
@@ -13,6 +16,7 @@ from fsspec.callbacks import DEFAULT_CALLBACK
 from fsspec.implementations.http import HTTPFileSystem
 from funcy import cached_property
 
+from scmrepo.git.backend.dulwich import _get_ssh_vendor
 from scmrepo.git.credentials import Credential, CredentialNotFoundError
 
 from .exceptions import LFSError
@@ -35,19 +39,12 @@ class LFSClient(AbstractContextManager):
     _SESSION_RETRIES = 5
     _SESSION_BACKOFF_FACTOR = 0.1
 
-    def __init__(
-        self,
-        url: str,
-        git_url: Optional[str] = None,
-        headers: Optional[dict[str, str]] = None,
-    ):
+    def __init__(self, url: str):
         """
         Args:
             url: LFS server URL.
         """
         self.url = url
-        self.git_url = git_url
-        self.headers: dict[str, str] = headers or {}
 
     def __exit__(self, *args, **kwargs):
         self.close()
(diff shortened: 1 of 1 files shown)


pmrowla:
> * Is using `_get_ssh_vendor` from `scmrepo.git.backend.dulwich` the right way to create an SSH client?

Ideally I think we should be using asyncssh directly so that this is independent of any particular git backend. I'm not sure how the CLI LFS client works with things like `GIT_SSH_COMMAND`, so it may actually be necessary to use dulwich if we do need to support falling back to a user-provided SSH shell command.

Either way, using dulwich here should be fine for now.

> * Any suggestions about testing? I don't feel too comfortable adding this feature without tests, but I'm not sure what the best approach to testing would be. Perhaps using a Git LFS test server? I've manually tested both Git SSH URLs and Git HTTP URLs of public and private projects, all scenarios are working for me.

I think testing this will really require setting up a docker container that is running a git + LFS + SSH server instance. We do already have a minimal setup for testing git over SSH, you can take a look at `tests/conftest.py` and `tests/docker-compose.yml` to get an idea of how that setup works.

---

In general this PR LGTM (other than the testing issue), but I should note that my understanding is that this use of LFS auth over SSH is not widely used, and the alternative (HTTPS auth) works for github/gitlab as well as huggingface. So I'm not sure this feature is something that needs to be prioritized.

@sisp is your team actively using LFS auth over SSH?

(Full disclosure, I'm leaving the DVC team at the end of this week, and at least in the short term, the rest of the team won't have the bandwidth to work on or review changes in the scmrepo LFS client other than bugs where something is completely broken with github/gitlab/huggingface)

sisp:
Thanks for your comprehensive feedback, @pmrowla! :pray: First of all, I'm sorry to hear you're leaving the DVC team. Your work on LFS support here – and thus also for DVC – has been fantastic and triggered an idea for a partial solution related to my struggles with tightly integrating DVC and GitLab (https://github.com/iterative/dvc-http/issues/50#issuecomment-1605291458, https://github.com/iterative/dvc-http/pull/56, [gitlab.com:gitlab-org/gitlab#413612](https://gitlab.com/gitlab-org/gitlab/-/issues/413612)) by simply using Git LFS for data/artifacts storage instead of a DVC remote. This is feasible on our corporate self-hosted GitLab instance, as it has a very generous LFS storage size limit, so storing significant data and artifacts is no problem. With DVC's LFS support, we can, e.g., import data managed via LFS from non-DVC repos and yet benefit from DVC's other features. That said, LFS is only a substitute for the regular cache but not the run-cache.

Since there's little time left before you leave, I'm inclined to leave the PR as is and look into testing in a follow-up. Also, I've looked into refactoring the current approach to using authenticator classes. I realized that not only retrieving credentials for the LFS Batch API is different for HTTP and SSH, but also constructing the LFS server URL from the Git URL is different. An authenticator class should not cover the URL construction, as that's a different concern, but authentication and URL construction are closely related. For this reason, I think we should keep the inheritance-based approach as is.

We use Git over SSH on our GitLab instance by default. For us, 2FA is always enabled, and then:

> When 2FA is enabled, you can’t use your password to authenticate with Git over HTTPS or the [GitLab API](https://docs.gitlab.com/ee/api/rest/index.html). You can use a [personal access token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) instead.
>
> &mdash; https://docs.gitlab.com/ee/user/profile/account/two_factor_authentication.html#use-personal-access-tokens-with-two-factor-authentication

The same applies to GitHub. In fact, I use 2FA and Git over SSH also on github.com and gitlab.com because it's secure and convenient. Git over HTTP only works nicely without 2FA or with public projects IMO, so I think supporting LFS for Git SSH URLs is a high priority in corporate environments, e.g. when not only consuming public LFS-managed files but also internal ones.

pmrowla (approved):

PR #340, opened 2024-03-13
lfs: add support for rate limit retries

sisp:
I've added support for handling rate limit responses and retrying requests accordingly.

Typical rate limit responses use the status code 429 and include one of the following headers:

* `Retry-After: <delay in seconds>` is used by, e.g., GitLab
* `RateLimit-Reset: <timestamp>` is used by, e.g., GitLab
* `X-RateLimit-Reset: <timestamp>` is used by, e.g., GitHub

The implementation is inspired by the [corresponding implementation in the `python-gitlab` library](https://github.com/python-gitlab/python-gitlab/blob/14a3ffe4cc161be51a39c204350b5cd45c602335/gitlab/client.py#L764-L775) and based on the [GitHub documentation on rate limits](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#checking-the-status-of-your-rate-limit).

### DIFF
--- a/src/scmrepo/git/lfs/client.py
+++ b/src/scmrepo/git/lfs/client.py
@@ -6,7 +6,9 @@ import shutil
 from abc import abstractmethod
 from collections.abc import Iterable, Iterator
 from contextlib import AbstractContextManager, contextmanager, suppress
+from http import HTTPStatus
 from tempfile import NamedTemporaryFile
+from time import time
 from typing import TYPE_CHECKING, Any, Optional
 
 import aiohttp
@@ -64,11 +66,12 @@ class LFSClient(AbstractContextManager):
                     sock_connect=self._REQUEST_TIMEOUT,
                     sock_read=self._REQUEST_TIMEOUT,
                 ),
-                retry_options=ExponentialRetry(
+                retry_options=_ExponentialRetry(
                     attempts=self._SESSION_RETRIES,
                     factor=self._SESSION_BACKOFF_FACTOR,
                     max_timeout=self._REQUEST_TIMEOUT,
                     exceptions={aiohttp.ClientError},
+                    statuses={HTTPStatus.TOO_MANY_REQUESTS},
                 ),
                 **kwargs,
             )
@@ -272,3 +275,18 @@ def _as_atomic(to_info: str, create_parents: bool = False) -> Iterator[str]:
         raise
     else:
         shutil.move(tmp_file.name, to_info)
+
+
+class _ExponentialRetry(ExponentialRetry):
+    def get_timeout(
+        self, attempt: int, response: Optional[aiohttp.ClientResponse] = None
+    ) -> float:
+        if response is not None and response.status == HTTPStatus.TOO_MANY_REQUESTS:
+            if "Retry-After" in response.headers:
+                with suppress(ValueError):
+                    return float(response.headers["Retry-After"])
+            for k in ["RateLimit-Reset", "X-RateLimit-Reset"]:
+                if k in response.headers:
+                    with suppress(ValueError):
+                        return float(response.headers[k]) - time()
+        return super().get_timeout(attempt, response)


skshetry:
@sisp, does this honor `attempts`?

sisp:
@skshetry Do you mean the `attempt` argument of `get_timeout()` method?

sisp:
Ah, now I see what you mean. If a request gets `Retry-After` on all 5 attempts, then no further attempts are made. That's because `get_timeout()` may only use the `attempt` counter, but the attempts are counted and managed outside this class in [`aiohttp_retry.client._RequestContext._do_request()`](https://github.com/inyutin/aiohttp_retry/blob/ba2169891f5b32a5c59e48ca185dd8e68e44ded7/aiohttp_retry/client.py#L104). So, it doesn't keep retrying beyond 5 in this case.

skshetry:
That's great. The only concern that I have with this PR is that a malicious server could return a large value, that might lead the user to believe `dvc` is frozen. 

So I am not sure if this should be the default, or should be enabled via some config.

skshetry:
How does `git-lfs` work?

src/scmrepo/git/lfs/client.py
```diff
@@ -272,3 +275,18 @@ def _as_atomic(to_info: str, create_parents: bool = False) -> Iterator[str]:
...
+
+
+class _ExponentialRetry(ExponentialRetry):
+    def get_timeout(
+        self, attempt: int, response: Optional[aiohttp.ClientResponse] = None
+    ) -> float:
+        if response is not None and response.status == HTTPStatus.TOO_MANY_REQUESTS:
+            if "Retry-After" in response.headers:
+                with suppress(ValueError):
+                    return float(response.headers["Retry-After"])
```
skshetry, line 287: Looks like `Retry-After` could be a timestamp too. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After.

skshetry: Never mind, let's leave it for now. Support for GitHub/GitLab may be good enough.

skshetry:
@pmrowla, if you have any thoughts? 👋🏽

skshetry:
> The only concern that I have with this PR is that a malicious server could return a large value, that might lead the user to believe dvc is frozen.

We do assume that the remotes are trusted by the users, so that may not be a big issue as I am making.

src/scmrepo/git/lfs/client.py
```diff
@@ -272,3 +275,18 @@ def _as_atomic(to_info: str, create_parents: bool = False) -> Iterator[str]:
...
+            for k in ["RateLimit-Reset", "X-RateLimit-Reset"]:
+                if k in response.headers:
+                    with suppress(ValueError):
+                        return float(response.headers[k]) - time()
```
shcheklein, line 291: just to make sure - is it always UTC in headers? 

it might also happen in rare cases that we get a negative number here - need to check the code above is ready for that

src/scmrepo/git/lfs/client.py
```diff
@@ -64,11 +66,12 @@ async def get_client(**kwargs):
                     sock_connect=self._REQUEST_TIMEOUT,
                     sock_read=self._REQUEST_TIMEOUT,
                 ),
-                retry_options=ExponentialRetry(
+                retry_options=_ExponentialRetry(
                     attempts=self._SESSION_RETRIES,
                     factor=self._SESSION_BACKOFF_FACTOR,
                     max_timeout=self._REQUEST_TIMEOUT,
                     exceptions={aiohttp.ClientError},
+                    statuses={HTTPStatus.TOO_MANY_REQUESTS},
```
shcheklein, line 74: what was the default for this? was it retrying on all codes?

shcheklein (commented):
Looks good to me, a few questions to clarify. Thanks @sisp !

Can we also add some tests for this please?

src/scmrepo/git/lfs/client.py
skshetry: Both gitlab/github return Unix time.

skshetry: I don't think we need to worry about negative times. Even if it does, it won't wait for that time, and should be fine.

src/scmrepo/git/lfs/client.py
sisp: The default value is `{}`, it was retrying only on responses with 5xx status codes.

* [`statuses` defaults to `None`](https://github.com/inyutin/aiohttp_retry/blob/ba2169891f5b32a5c59e48ca185dd8e68e44ded7/aiohttp_retry/retry_options.py#L45) and [corresponds to an empty set](https://github.com/inyutin/aiohttp_retry/blob/ba2169891f5b32a5c59e48ca185dd8e68e44ded7/aiohttp_retry/retry_options.py#L22-L23).
* [Requests are retried when the response status code is a member of the `statuses` set](https://github.com/inyutin/aiohttp_retry/blob/ba2169891f5b32a5c59e48ca185dd8e68e44ded7/aiohttp_retry/client.py#L90-L91).
* [Requests with 5xx response status codes are retried when `retry_all_server_errors` is `True`](https://github.com/inyutin/aiohttp_retry/blob/ba2169891f5b32a5c59e48ca185dd8e68e44ded7/aiohttp_retry/client.py#L87-L88), which is the [default value](https://github.com/inyutin/aiohttp_retry/blob/ba2169891f5b32a5c59e48ca185dd8e68e44ded7/aiohttp_retry/retry_options.py#L47).

sisp:
@shcheklein There aren't any integration tests against an LFS test server yet, so testing rate limit retries would mean significant test setup work first. I'd defer this to follow-up PRs. WDYT?

src/scmrepo/git/lfs/client.py
shcheklein: sounds good, folks!

src/scmrepo/git/lfs/client.py
shcheklein: do, just to double check - it will keep retrying on 5xx the same way as we before, we only add new codes to retry on, right?

this kind of things would be nice to have tests for (mocks, not integration)

sisp: Yes, that's correct.

shcheklein (commented):
Your call on unit tests. Thanks @sisp !

shcheklein (approved):

sisp:
I'll look into adding unit tests tomorrow.

shcheklein:
@sisp you also should be able to merge PRs now (I've added you as collaborator to the project).

sisp:
Wow, thanks, I appreciate your trust. :pray:

pmrowla (approved):

PR #341, opened 2024-03-15
lfs: add tests for rate limit retries

sisp:
I've added some tests for rate limit retries.

Follow-up of #340.

/cc @shcheklein

### DIFF
--- a/tests/test_lfs.py
+++ b/tests/test_lfs.py
@@ -102,36 +102,26 @@ class CallbackResultRecorder:
         return self._results[url]
 
 
-@pytest.mark.parametrize(
-    "rate_limit_header",
-    [
-        lambda: {"Retry-After": "1"},
-        lambda: {"RateLimit-Reset": f"{int(time()) + 1}"},
-        lambda: {"X-RateLimit-Reset": f"{int(time()) + 1}"},
-    ],
-)
-def test_rate_limit_retry(
-    storage: LFSStorage, rate_limit_header: Callable[[], dict[str, str]]
-):
-    client = LFSClient.from_git_url("http://git.example.com/namespace/project.git")
-    lfs_batch_url = f"{client.url}/objects/batch"
-    lfs_object_url = f"http://git-lfs.example.com/{FOO_OID}"
-    recorder = CallbackResultRecorder()
+class LFSServerMock:
+    def __init__(
+        self,
+        mocker: aioresponses,
+        recorder: CallbackResultRecorder,
+        batch_url: str,
+        objects_url: str,
+    ) -> None:
+        self._mocker = mocker
+        self._recorder = recorder
+        self.batch_url = batch_url
+        self.objects_url = objects_url
 
-    with aioresponses() as m:
-        m.post(
-            lfs_batch_url,
-            callback=recorder.record(
-                CallbackResult(
-                    status=HTTPStatus.TOO_MANY_REQUESTS,
-                    headers=rate_limit_header(),
-                    reason="Too many requests",
-                )
-            ),
-        )
-        m.post(
-            lfs_batch_url,
-            callback=recorder.record(
+    def get_object_url(self, oid: str) -> str:
+        return f"{self.objects_url}/{oid}"
+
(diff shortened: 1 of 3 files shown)


tests/test_lfs.py
```diff
@@ -74,3 +85,197 @@ def test_lfs(tmp_dir: TmpDir, scm: Git, lfs_objects: TmpDir):
...
+
+@pytest.mark.parametrize(
+    "rate_limit_header",
+    [
+        lambda: {"Retry-After": "1"},
+        lambda: {"RateLimit-Reset": f"{int(time()) + 1}"},
+        lambda: {"X-RateLimit-Reset": f"{int(time()) + 1}"},
+    ],
+)
+def test_rate_limit_max_retries_batch(
```
shcheklein, line 194: it seems this already covered by the test above?

tests/test_lfs.py
```diff
@@ -74,3 +85,197 @@ def test_lfs(tmp_dir: TmpDir, scm: Git, lfs_objects: TmpDir):
...
+                    reason="Too many requests",
+                ),
+            ),
+            repeat=True,
+        )
+
+        with pytest.raises(ClientResponseError, match="Too many requests"):
+            client.download(storage, [Pointer(oid=FOO_OID, size=FOO_SIZE)])
+
+        results = recorder[lfs_batch_url]
```
shcheklein, line 277: can we simplify the test and use `lfs_batch_url` to test it, instead of `lfs_object_url`?

tests/test_lfs.py
sisp: Do you mean the parametrization with the different rate limit headers? Or what specifically do you think is already covered? 🤔 

This test ensures that rate limit retries for the batch API don't continue indefinitely.

tests/test_lfs.py
sisp: I'm not sure I understand your proposal. Here, I'd like to make sure that rate limit retries work for the object API.

tests/test_lfs.py
shcheklein: okay, yes, I see the difference 👍

tests/test_lfs.py
shcheklein: yes, I see the difference now

tests/test_lfs.py
```diff
@@ -74,3 +85,197 @@ def test_lfs(tmp_dir: TmpDir, scm: Git, lfs_objects: TmpDir):
...
+def test_rate_limit_max_retries_objects(
+    storage: LFSStorage, rate_limit_header: Callable[[], dict[str, str]]
+):
+    client = LFSClient.from_git_url("http://git.example.com/namespace/project.git")
+    lfs_batch_url = f"{client.url}/objects/batch"
+    lfs_object_url = f"http://git-lfs.example.com/{FOO_OID}"
+    recorder = CallbackResultRecorder()
+
+    with aioresponses() as m:
+        m.post(
```
shcheklein, line 237: ok, now that I see the difference between tests, can we move each of these `m.post`, `m.get` blocks into a separate function and reuse them across tests? so that each test is small in size (and probably will be move descriptive)? wdyt?

sisp: Good idea! :+1: I've pushed a commit with refactored tests. WDYT?

shcheklein (approved):
thanks @sisp ! good stuff. Feel free to merge (we usually do squash merge I think) and probably you should be able to trigger the new release (releases -> create a new release  I think). We can then bump the version required on the DVC side if you feel it is needed.

sisp:
Thanks for the helpful review, @shcheklein! 🙇 

IMO, no release is needed, as the only change since the last one is the added tests by this PR. [v3.3.0](https://github.com/iterative/scmrepo/releases/tag/3.3.0) already includes the feature. No need for a bump in DVC from my side, [`scmrepo>=3,<4`](https://github.com/iterative/dvc/blob/017a510b83a682ddf4059a3d33bb32bc63324bba/pyproject.toml#L63) includes v3.3.0.

PR #345, opened 2024-03-19
unpin test deps

skshetry:


sisp:
Just out of curiosity: Why do you think unpinning dev/test dependencies is a good idea? I'd argue in favor of pinning – and even locking the entire dependency tree – and letting Dependabot or Renovate help in maintaining the pins/locks. Without pinning/locking, dev environments are non-deterministic, so tracing bugs becomes more difficult across environments. In fact, [Renovate's `config:best-practices` preset](https://docs.renovatebot.com/presets-config/#configbest-practices) includes the [`:pinDevDependencies` preset](https://docs.renovatebot.com/presets-default/#pindevdependencies). IMO, [`deptry`](https://github.com/fpgmaas/deptry) has a great setup w.r.t. dependency management.

WDYT, @skshetry?

sisp:
Thanks for clarifying and giving some background info. :+1: Have you considered using [Renovate with auto-merge](https://docs.renovatebot.com/key-concepts/automerge/)? `deptry` also uses it for some dependency types. If you're unpinning dependencies, there's no downside to automerging as you're not reviewing those changes either way. Just my 2 cents. :wink:

PR #355, opened 2024-05-03
lfs: optimize path filtering

### LINKED ISSUE #338: lfs: internal `_filter_paths()` function is prohibitively slow
The internal helper function [`_filter_paths()`](https://github.com/iterative/scmrepo/blob/fb99c42816f3c58a7d379092479d72338c3c2cac/src/scmrepo/git/lfs/fetch.py#L121-L133) in the `lfs` subpackage is prohibitively slow at the moment.

For instance, when I host the (rather small) MNIST dataset (70,000 image files) using Git LFS and import it using `dvc import`, then [`_filter_paths()` is effectively called](https://github.com/iterative/scmrepo/blob/fb99c42816f3c58a7d379092479d72338c3c2cac/src/scmrepo/git/lfs/fetch.py#L108) like this:

```python
paths = ["data/train/0/1.png", "data/train/0/2.png", ...]  # list of all 70,000 image files

for path in _filter_paths(paths, include=paths, exclude=None):
    # yield LFS pointer data
```

So, in this simple case, the current implementation has quadratic complexity in the number of LFS files and materializes generators unnecessarily, which is already significant with only 70,000 files – imagine large datasets.

I've done a quick benchmark to quantify the current implementation performance and compare it against two alternatives:

* Current implementation:

    ```python
    import timeit

    setup = """\
    import fnmatch
    from typing import Iterable, Iterator, Optional

    import fsspec


    def _filter_paths(
        paths: Iterable[str], include: Optional[list[str]], exclude: Optional[list[str]]
    ) -> Iterator[str]:
        filtered = set()
        if include:
            for pattern in include:
                filtered.update(fnmatch.filter(paths, pattern))
        else:
            filtered.update(paths)
        if exclude:
            for pattern in exclude:
                filtered.difference_update(fnmatch.filter(paths, pattern))
        yield from filtered


    fs = fsspec.filesystem("file")
    paths = fs.find("data/")
    """


    timeit.timeit("list(_filter_paths(paths, include=paths, exclude=None))", setup=setup, number=1)
    # >>> 632.4136102720004 seconds
    ```

* A more efficient implementation with identical semantics by manually unionizing the filename regex patterns and matching each path against the pre-compiled single regex:

    ```python
    import timeit

    setup = """\
    import fnmatch
    import re
    from typing import Iterable, Iterator, Optional

    import fsspec


    def _filter_paths(
        paths: Iterable[str], include: Optional[list[str]], exclude: Optional[list[str]]
    ) -> Iterator[str]:
        if include:
            include_match = re.compile(r"|".join(map(fnmatch.translate, include))).match
            paths = (path for path in paths if include_match(path) is not None)
        if exclude:
            exclude_match = re.compile(r"|".join(map(fnmatch.translate, exclude))).match
            paths = (path for path in paths if exclude_match(path) is None)
        yield from paths


    fs = fsspec.filesystem("file")
    paths = fs.find("data/")
    """


    timeit.timeit("list(_filter_paths(paths, include=paths, exclude=None))", setup=setup, number=1)
    # >>> 118.97085808499833 seconds
    ```

* Assuming we actually don't need Unix filename pattern matching (`fnmatch`) – it [doesn't seem to be needed _at the moment_ by DVC](https://github.com/iterative/dvc/blob/7df58ce404d682e7fd8518374648428d171ca159/dvc/scm.py#L280-L285) –, an implementation using set lookups (O(1) complexity) would be _orders of magnitude_ faster:

    ```python
    import timeit

    setup = """\
    import fnmatch
    import re
    from typing import Iterable, Iterator, Optional

    import fsspec


    def _filter_paths(
        paths: Iterable[str], include: Optional[list[str]], exclude: Optional[list[str]]
    ) -> Iterator[str]:
        if include:
            include_set = set(include)
            paths = (path for path in paths if path in include_set)
        if exclude:
            exclude_set = set(exclude)
            paths = (path for path in paths if path not in exclude_set)
        yield from paths


    fs = fsspec.filesystem("file")
    paths = fs.find("data/")
    """


    timeit.timeit("list(_filter_paths(paths, include=paths, exclude=None))", setup=setup, number=1)
    # >>> 0.00669500499861897 seconds
    ```

WDYT about the two alternative implementations? Is Unix filename pattern matching really needed?

sisp:
I've optimized LFS path filtering as we discussed in #338. The first optimization implements the suggestion in https://github.com/iterative/scmrepo/issues/338#issuecomment-2081256097 to short-cut a single-path `include` filter. Here are the tests for the regex that extracts the path prefix: https://regex101.com/r/wBjHf0/1 Note that the extra `\n` on regex101.com is only necessary to allow one test case per line, it isn't needed in the actual regex. ~~The second optimization unionizes the filename regex patterns derived from Unix filename patterns and matches each path against the pre-compiled single regex, which is faster than matching against the Unix filename patterns individually. Also, it avoids intermediate list materialization but instead implements a streaming filter.~~

~~As the two commits implement independent optimizations, I intend this PR to be rebase-merged without commit squashing.~~

Partially fixes #338.

### DIFF
--- a/src/scmrepo/git/lfs/fetch.py
+++ b/src/scmrepo/git/lfs/fetch.py
@@ -1,6 +1,7 @@
 import fnmatch
 import io
 import os
+import re
 from collections.abc import Iterable, Iterator
 from typing import TYPE_CHECKING, Callable, Optional
 
@@ -98,6 +99,9 @@ def get_fetch_url(scm: "Git", remote: Optional[str] = None):  # noqa: C901,PLR09
     return scm.get_remote_url(remote)
 
 
+_ROOT_PATH_PREFIX_REGEX = re.compile(r"^(?P<prefix>[^*?\[]*(?:/|$))")
+
+
 def _collect_objects(
     scm: "Git",
     rev: str,
@@ -105,7 +109,25 @@ def _collect_objects(
     exclude: Optional[list[str]],
 ) -> Iterator[Pointer]:
     fs = scm.get_fs(rev)
-    for path in _filter_paths(fs.find("/"), include, exclude):
+    # Optimize path filtering if the `include` list contains exactly one path.
+    # First, determine the root directory wherein to initiate the file search.
+    # If the `include` path is a Unix filename pattern, determine the static
+    # path prefix and set it as the root directory. Second, if the path and the
+    # root are identical or the Unix filename pattern matches *any* (i.e., `**`)
+    # file under the root directory, unset `include` to avoid unnecessary
+    # filtering work.
+    if (
+        include
+        and len(include) == 1
+        and (result := _ROOT_PATH_PREFIX_REGEX.match(path := include[0]))
+    ):
+        root = result.group("prefix")
+        if path in {root, f'{root.rstrip("/")}/**'}:
+            include = []
+    else:
+        root = "/"
+
+    for path in _filter_paths(fs.find(root), include, exclude):
         check_path = path.lstrip("/")
         if scm.check_attr(check_path, "filter", source=rev) == "lfs":
             try:


src/scmrepo/git/lfs/fetch.py
```diff
@@ -98,14 +99,35 @@ def get_fetch_url(scm: "Git", remote: Optional[str] = None):  # noqa: C901,PLR09
...
+    # First, determine the root directory wherein to initiate the file search.
+    # If the `include` path is a Unix filename pattern, determine the static
+    # path prefix and set it as the root directory. Second, if the path and the
+    # root are identical or the Unix filename pattern matches *any* (i.e., `**`)
+    # file under the root directory, unset `include` to avoid unnecessary
+    # filtering work.
+    if (
+        include
+        and len(include) == 1
+        and (result := _FNPATTERN_PREFIX_REGEX.match(path := include[0]))
```
shcheklein, line 122: QQ: what is FNPATTERN (why this name?)

src/scmrepo/git/lfs/fetch.py
```diff
@@ -121,16 +143,13 @@ def _collect_objects(
 def _filter_paths(
     paths: Iterable[str], include: Optional[list[str]], exclude: Optional[list[str]]
 ) -> Iterator[str]:
-    filtered = set()
     if include:
-        for pattern in include:
-            filtered.update(fnmatch.filter(paths, pattern))
-    else:
-        filtered.update(paths)
+        include_match = re.compile(r"|".join(map(fnmatch.translate, include))).match
```
shcheklein, line 147: let's do this a separate PR please? also, if we have a lot of patterns (100s) - will it handle it fine?

src/scmrepo/git/lfs/fetch.py
```diff
@@ -98,14 +99,35 @@ def get_fetch_url(scm: "Git", remote: Optional[str] = None):  # noqa: C901,PLR09
...
+    # root are identical or the Unix filename pattern matches *any* (i.e., `**`)
+    # file under the root directory, unset `include` to avoid unnecessary
+    # filtering work.
+    if (
+        include
+        and len(include) == 1
+        and (result := _FNPATTERN_PREFIX_REGEX.match(path := include[0]))
+    ):
+        root = result.group("prefix")
+        if path in {root, f'{root.rstrip("/")}/**'}:
```
shcheklein, line 125: should we do rstrip if "root != '/'" at the very beginning - to mormalize it once and "forget"

src/scmrepo/git/lfs/fetch.py
shcheklein: okay, I see it's fnmatch. It's fine, super minor - can be probably _ROOT_PATH_PREFIX_REGEX ...

src/scmrepo/git/lfs/fetch.py
```diff
@@ -98,14 +99,35 @@ def get_fetch_url(scm: "Git", remote: Optional[str] = None):  # noqa: C901,PLR09
     return scm.get_remote_url(remote)
 
 
+_FNPATTERN_PREFIX_REGEX = re.compile(r"^(?P<prefix>[^*?\[]*(?:/|$))")
```
shcheklein, line 102: `[^*?\[]]` - should it include `!` and `]`?

to double check - would it match - /root/multiple/directories ?

(there is a bug in the existing implementation I think btw - if path contains some of those special symbols - collect_objects will probably, break or behave in an unpredictable way)

src/scmrepo/git/lfs/fetch.py
sisp: Sounds better. :+1: I've amended the commit to use this name.

src/scmrepo/git/lfs/fetch.py
sisp: Agreed, I've dropped the commit here and created submitted a separate PR at #356.

The unionized regex can easily handle 100s – even orders of magnitude more – patterns without any problems. Check this example:

```python
import fnmatch
import re

re.compile(r"|".join(map(fnmatch.translate, [f"/a/b/{i}/**" for i in range(100000)]))).match("/a/b/1/")
# <re.Match object; span=(0, 7), match='/a/b/1/'>
```

Also note that unionizing the derived regexes is an [undocumented but known technique](https://github.com/python/cpython/blob/11594da0461b02a8448dfa3354884ef6e274e587/Lib/fnmatch.py#L163-L165).

src/scmrepo/git/lfs/fetch.py
sisp: > `[^*?\[]]` - should it include `!` and `]`?

I think it's fine to only match `[` because `!` and `]` must come after `[`, otherwise those characters are not used in to create a Unix filename pattern.

> to double check - would it match - /root/multiple/directories ?

Yes, see https://regex101.com/r/wBjHf0/2 where I've added this case at the bottom.

> (there is a bug in the existing implementation I think btw - if path contains some of those special symbols - collect_objects will probably, break or behave in an unpredictable way)

Do you mean that a non-pattern/literal path name in `include` such as `a/b/c[!x]/?/*/**' would incorrectly be interpreted as a Unix filename pattern? For instance,

```python
from fnmatch import fnmatch

fnmatch("a/b/cd/e/f/g", "a/b/c[!x]/?/*/**")
# True
```

should return `False` if the second argument of `fnmatch()` is actually a non-pattern/literal path name. A solution to the problem could be this: We consider the items in the `include` and `exclude` lists _always_ as Unix filename patterns and _never_ as literal paths. If somebody would like to pass a literal path (e.g. [DVC's LFS prefetching](https://github.com/iterative/dvc/blob/a7c0f6ef398dfd90ffabe8ed5edce2237c473b36/dvc/fs/__init__.py#L72-L74) call), then the literal path (component) must be escaped using [`glob.escape()`](https://docs.python.org/3/library/glob.html#glob.escape). For DVC, I'd apply the following fix in [`dvc/fs/__init__.py`](https://github.com/iterative/dvc/blob/a7c0f6ef398dfd90ffabe8ed5edce2237c473b36/dvc/fs/__init__.py#L72-L74):

```diff
             lfs_prefetch(
-                fs, [f"{fs.normpath(fs_path)}/**" if fs.isdir(fs_path) else fs_path]
+                fs, [f"{glob.escape(fs.normpath(fs_path))}/**" if fs.isdir(fs_path) else fs_path]
             )
```

WDYT?

src/scmrepo/git/lfs/fetch.py
shcheklein: yep, in this case I'm still a bit worried bc it could explode (we don't really control the number of files) - also, I don't know about the performance implications (how much time compile takes, if we run each file by this regex - is it really better vs running one by one). Anyways, we can continue the discussion in the new PR. Thanks for addressing this.!

src/scmrepo/git/lfs/fetch.py
shcheklein: thanks!

src/scmrepo/git/lfs/fetch.py
shcheklein: > I think it's fine to only match [ because ! and ] must come after [, otherwise those characters are not used to create a Unix filename pattern.

probably, yep. Comes down to how we treat paths like this: `/roo]t/something/**` and include / exclude in general. With the fix that you are suggesting below, will it still work (a path that contains some escaped special symbols). Or will we in that case just ignore the optimizations?

Is it safer to ignore it even in case like `/roo]t/something/**`?

> A solution to the problem could be this ...

yep, sounds good to me!

sisp: Incomplete syntax for patterns is ignored by the escape function:

```python
import glob

glob.escape("/roo]t/something/**")
# '/roo]t/something/[*][*]'
```

Thus, the optimization will treat `/roo]t/something/` as the root directory. I think it's safe to apply my suggested fix to DVC; I'll submit a PR for you to review.

sisp: See https://github.com/iterative/dvc/pull/10416.

shcheklein (approved):

PR #360, opened 2024-05-20
replace deprecated `GIT_*` constants with `pygit2.enums`

skshetry:
This was deprecated in 1.14.0. See https://github.com/libgit2/pygit2/blob/facaf22db399c6ec429df62f827d37f11a6472e6/CHANGELOG.md?plain=1#L114.


shcheklein (approved):

dberenbaum:
Thanks @skshetry! Do we need to bump the min version in dvc?

PR #369, opened 2024-07-08
Handle missing password

dberenbaum:
Empty passwords were being passed as `()` instead of `None` to asyncssh, causing unexpected failure.

See https://discuss.dvc.org/t/error-in-pushing-experiments/2138 and the added test for how to reproduce it.

In https://github.com/ronf/asyncssh/blob/050e5c1a0f42d5f145b44c33b183ff745447a687/asyncssh/connection.py#L7790, you can see that `username` defaults to `()`, but `password` defaults to `None`.


skshetry (dismissed):

skshetry:
@dberenbaum, please fix lint errors. Code looks good to me.

dberenbaum:
Linting is fixed now. PTAL @skshetry, thanks!

dberenbaum:
I think the force push dismissed your approval @skshetry

PR #378, opened 2024-09-26
pass and init proxy settings in dulwich and pygit2

shcheklein:
Fixes https://github.com/iterative/dvc/issues/10563

+ a few other occurrences in dulwich


shcheklein:
@skshetry should be good to go. Merging it, but please review when you have some time.

PR #404, opened 2025-02-06
remove asyncssh patching

skshetry:
Looks like this is no longer necessary, GitHub seems to return proper algorithm name these days, at least it does for me.

I tried with the following script:

```python
import asyncio

from asyncssh import ProtocolError, connect
from asyncssh.auth import MSG_USERAUTH_PK_OK, _ClientPublicKeyAuth


def _process(self, _pkttype, _pktid, packet):
    algorithm = packet.get_string()
    key_data = packet.get_string()
    packet.check_end()

    print(algorithm, self._keypair.algorithm)

    if algorithm != self._keypair.algorithm or key_data != self._keypair.public_data:
        raise ProtocolError("Key mismatch")

    self.create_task(self._send_signed_request())
    return True


_ClientPublicKeyAuth._packet_handlers[MSG_USERAUTH_PK_OK] = _process


async def func():
    async with connect(host="github.com", username="git", known_hosts=None):
        pass


asyncio.run(func())
```

```
b'rsa-sha2-256' b'rsa-sha2-256'
```

See https://github.com/iterative/dvc/pull/6983 - specially https://github.com/iterative/dvc/commit/25dca0b6653ff26f593dd85febf3366afd552f59 for context.

(scmrepo commit that was split from dvc: https://github.com/iterative/scmrepo/commit/2978b52180f15bc1bbd60e2809368f75d86108b8)

### DIFF
--- a/src/scmrepo/git/backend/dulwich/asyncssh_vendor.py
+++ b/src/scmrepo/git/backend/dulwich/asyncssh_vendor.py
@@ -97,45 +97,6 @@ class AsyncSSHWrapper(BaseAsyncObject):
     close = sync_wrapper(_close)
 
 
-# NOTE: Github's SSH server does not strictly comply with the SSH protocol.
-# When validating a public key using the rsa-sha2-256 or rsa-sha2-512
-# signature algorithms, RFC4252 + RFC8332 state that the server should respond
-# with the same algorithm in SSH_MSG_USERAUTH_PK_OK. Github's server always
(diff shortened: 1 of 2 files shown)


skshetry:
cc @pmrowla

shcheklein (approved):

pmrowla:
Should be fine to remove this now

skshetry:
Thank you, @pmrowla.

PR #419, opened 2025-07-11
git: implement list_all_commits using pygit2

skshetry:


shcheklein:
@skshetry rebased, just to see if tests pass

PR #421, opened 2025-07-17
fix dulwich warning ls-remote result dict

shcheklein:
Fixes a warning we have in tests:

```python
Warning: Treating LsRemoteResult as a dictionary is deprecated. Use result.refs instead.
```

due to:

https://github.com/jelmer/dulwich/blob/3c0213350d16e6909aef19f7eb54ab7c4ed2bafb/dulwich/client.py#L420

and

https://github.com/jelmer/dulwich/blob/3c0213350d16e6909aef19f7eb54ab7c4ed2bafb/dulwich/client.py#L321

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ requires-python = ">=3.9"
 dynamic = ["version"]
 dependencies = [
     "gitpython>3",
-    "dulwich>=0.22.1",
+    "dulwich>=0.23.1",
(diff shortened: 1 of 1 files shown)


skshetry:
This requires bumping minimum version of dulwich to [dulwich-0.23.1](https://github.com/jelmer/dulwich/releases/tag/dulwich-0.23.1).

skshetry (approved):

PR #431, opened 2025-07-31
dulwich: do not add files from submodules

skshetry:
`git` doesn't add files from submodules when staging, so we align `dulwich` with that behavior.

Also, `libgit2`/`pygit2` removes files from submodules on `reset()`, so raise `NotImplementedError` for now to avoid accidental deletions, and redirect such cases to the `GitPython` backend instead.

Fixes https://github.com/iterative/dvc/issues/10823.

### DIFF
--- a/src/scmrepo/git/backend/dulwich/__init__.py
+++ b/src/scmrepo/git/backend/dulwich/__init__.py
@@ -199,6 +199,9 @@ class DulwichBackend(BaseGitBackend):  # pylint:disable=abstract-method
         self._submodules: dict[str, str] = self._find_submodules()
         self._stashes: dict = {}
 
+    def list_submodules(self) -> list[str]:
+        raise NotImplementedError
+
     def _find_submodules(self) -> dict[str, str]:
         """Return dict mapping submodule names to submodule paths.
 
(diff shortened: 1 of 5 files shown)


src/scmrepo/git/backend/dulwich/__init__.py
```diff
@@ -356,14 +356,12 @@ def _expand_paths(self, paths: list[str], force: bool = False) -> Iterator[str]:
                 # that the result of relpath(path, root_dir) is actually the
                 # path relative to the submodule root.
                 fs_path = relpath(path, self.root_dir)
+                in_submodule = False
```
skshetry, line 359: @pmrowla, do you happen to know why we have this support for adding submodule files, instead of skipping them?


In `git`, I see the following:

```console
$ git add submodules/foo
fatal: Pathspec 'external/foo' is in submodule 'external'

$ git add submodules
# no changes
```

skshetry: > It might have been a case where a user had the DVC repo inside a submodule but was trying to run from the outer git repo?

Do we even support running from outside of the DVC repo? On the issue linked in that PR (https://github.com/iterative/dvc/issues/4969), I only see mention of subrepo.

pmrowla: No, but I think that issue probably predates whenever we made the formal decision to not support it. I tried looking for the original discord context but discord search is really bad for finding something that old.

I think your changes here are fine, I think there's a 99% chance my original change here was just for a specific user who asked for it 5 years ago, and it's almost certainly not worth maintaining now.

